PR #4, opened 2023-02-01
feat(config): add ability to configure widget outside of Widget instantiation

cerebrl:
## Summary

Originally, the Login Widget class instantiation was when the underlying SDK was configured. This prevented key methods from being used separately prior to the Widget itself from being used. This change allows for the configuration to set separate from the Login Widget.

## Details

The underlying SDK can now be configured outside of the Login Widget instantiation. This means the `config` prop of the Widget class is now optional. This PR also includes quite a bit of refactoring at the widget component level. The two form factors (inline and modal) now share the same API generation and logic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,40 @@ journey.onSuccess((response) => {
 
 And, that's it. You now can mount, display, and authenticate users through the ForgeRock Login Widget. There are addition features documented below for a more complete implementation. For more about Widget events, [see the Widget Events section](#widget-events).
 
+## Checking user information
+
+There are many reasons why you may need user information throughout your application. It may be to display the user's first name or email, it may be to use tokens for protected resources, or to validate an Access Token with the ForgeRock server. Any of these will require the underlying SDK to be configured. To do this, you have two choices:
+
+1. Configure the SDK through `Widget` instantiation
+2. Configure the SDK through the `configuration` object
+
+We've seen [configuring the SDK through `Widget` instantiation](#instantiate-the-widget-modal), but that may not always be preferable as the Widget may not be relevant in all parts of your app. Let's see how we can use the second option instead.
+
+```js
+let userInfo;
+
+// Configure the underlying SDK, so it can communicate with the ForgeRock platform
+configuration.set({
+  clientId: 'WebOAuthClient',
+  redirectUri: `${window.location.origin}/callback`,
+  scope: 'openid profile email me.read',
+  serverConfig: {
+    baseUrl: 'https://example.forgeblocks.com/am/',
+    timeout: 5000,
+  },
+  realmPath: 'alpha',
+});
+
+// Do we have stored tokens
+if (await user.tokens()) {
+  // User has tokens, so attempt call to /userinfo
+  userInfo = await user.info(true);
+} else {
+  // If no tokens, assign null
+  userInfo = null;
+}
+```
+
 ## Complete Widget API
 
 The Widget comes with methods and event handlers used to control the lifecycle of user journeys/authentication.
@@ -391,6 +425,65 @@ widget.$destroy();
 NOTE: For more SDK configuration options, please [see our SDK's configuration document](https://backstage.forgerock.com/docs/sdks/3.3/javascript/configuring/configuring-forgerock-sdk-settings-for-your-javascript-app.html), or you can [see our API docs for more developer detail](https://backstage.forgerock.com/docs/sdks/3.3/_attachments/javascript/api-reference-core/interfaces/configoptions.html).
 
 NOTE: For content schema, please [use the example en-US locale file](/src/locales/us/en/index.ts).
+
+### Configuration
+
+This object can be useful when you want to leverage APIs that require interaction with the ForgeRock platform or access stored tokens in a location of your app that may not directly use the `Widget`. When the `configuration.set` method is used, it removes the need to set the `config` property within the `Widget` class instantiation.
+
+For example, these would be equivalent:
+
+```js
+// Configuration within Widget instantiation
+const widget = new Widget({
+  target: document.getElementById('widget-root'),
+  props: {
+    config: {
+      /**
+       * REQUIRED; SDK configuration object
+       */
+      serverConfig: {
+        baseUrl: 'https://customer.forgeblocks.com/am',
+      },
+      /**
+       * OPTIONAL, *BUT ENCOURAGED*, CONFIGURATION
+       * Remaining config is optional with fallback values shown
+       */
+      clientId: 'WebLoginWidgetClient',
+      realmPath: 'alpha',
+      redirectUri: window.location.href,
+      scope: 'openid email',
+      tree: 'Login',
+    },
+  },
+});
+```
+
+```js
+// Configuration outside Widget instantiation
+configuration.set({
+  /**
+   * REQUIRED; SDK configuration object
+   */
+  serverConfig: {
+    baseUrl: 'https://customer.forgeblocks.com/am',
+  },
+  /**
+   * OPTIONAL, *BUT ENCOURAGED*, CONFIGURATION
+   * Remaining config is optional with fallback values shown
+   */
+  clientId: 'WebLoginWidgetClient',
+  realmPath: 'alpha',
+  redirectUri: window.location.href,
+  scope: 'openid email',
+  tree: 'Login',
+});
+
+const widget = new Widget({
+  target: document.getElementById('widget-root'),
+});
+```
+
+Note: It's important to not that all methods of the `user` API imported from this module require configuration to be set. So, setting the configuration is best done at the top index or entry file of your application. Then you can use the all APIs without failure.
 
 ### Journey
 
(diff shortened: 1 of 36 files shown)


src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -0,0 +1,232 @@
...
+  Config,
+  FRUser,
+  SessionManager,
+  TokenManager,
+  UserManager,
+  type GetTokensOptions,
+} from '@forgerock/javascript-sdk';
+import { get } from 'svelte/store';
+
+export function widgetApiFactory(modal: Modal) {
```
ryanbas21, line 24: I'm waffling on how i feel about this. I kind of think this should be a `class`. However if we want to stay "functional" we can probably use some other tools from that realm. 

I am hesitant to have a stateful factory, it just feels odd to me. This seems like a good use of a state machine though. Have we considered this as an option? 

Here are ways I see us building this without what I feel is a "stateful" factory since we have to keep the response value in the factory function outside of the configuration object.

1. a class, as this is more or less what we've built here.
2. we can use something _like_ a state monad (sorry for monad word).
3. a state machine (like xstate type implementation).
4. We can keep this implementation.

this implementation feels just an awful lot like a class, so i'm questioning why even take this approach versus a class?

To be clear, there isn't anything inherently wrong with the code. However, if the intent of the code is "to be functional" I think we are looking at what is equivalent to a `class` without the `new` keyword.

cerebrl: I had the same thoughts as well. I actually started writing a class implementation of this. I had a few issues with it after some experimentation. Here they are in bullet form, but I can go into more details, if desired:

What I wanted to avoid ...

1. prototype pollution or misc. state hanging into the exported API
2. worrying about the `this` reference
3. worrying about method versus function execution context
4. leaking state information
5. cumbersome object shape influenced by the class pattern

What I did want ...

1. be able to export the created object as-is for a clean API
2. object methods that have a persistent, explicit state reference, rather than an implicit one
3. no-nonsense, hierarchical object creation free from any restriction or hacking

It's worth noting that the block of code within the `<script type="module">` annotation is only run once when the app bootstraps itself. It is not rerun unless the browser is reloaded. Reference: https://svelte.dev/docs#component-format-script-context-module.

Now, with that said, I'm completely open to the use of classes if you feel it's still the best way forward, but I would like to ask for your help. I am terrible at writing classes, and the class feature itself may have improved since I last used them, so your help in refactoring this to use them would be awesome.

src/lib/widget/inline.svelte
```diff
@@ -191,41 +49,43 @@
...
   /**
    * Initialize the stores and ensure both variables point to the same reference.
    * Variables with _ are the reactive version of the original variable from above.
    */
-  let _journeyStore = (journeyStore = initializeJourney(config));
-  let _oauthStore = (oauthStore = initializeOauth(config));
-  let _userStore = (userStore = initializeUser(config));
+  api.setJourneyStore(initializeJourney(config));
+  api.setOAuthStore(initializeOauth(config));
+  api.setUserStore(initializeUser(config));
```
cerebrl, line 81: Now that we are using this `api` object created from the factory, we don't need the double references between the `script` here and the one (`context="module"`) from above. Using an object instead allows for passing the needed stores by reference, rather than value.

src/routes/e2e/widget/+page.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
```
cerebrl, line 1: I added this file as a "page" that has no widget, just simple user info request. I think I'll add a bit more here, but the general intent is to have a route that doesn't have the widget, but still doesn't use the underlying SDK. This tests the configuration feature.

src/routes/e2e/widget/+page.svelte
```diff
@@ -0,0 +1,61 @@
...
+    });
+
+    // Fetch tokens locally to see if the user has tokens
+    if (await user.tokens()) {
+      fetchedTokens = true;
+      userInfo = await user.info(true);
+    } else {
+      // Set fetched tokens to true so the links render to login
+      fetchedTokens = true;
+    }
```
cerebrl, line 33: Added just a bit more here to test the other APIs.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -5,6 +5,13 @@ import { asyncEvents, verifyUserInfo } from '../../utilities/async-events.js';
 test('Modal widget with login', async ({ page }) => {
   const { clickButton, navigate } = asyncEvents(page);
 
+  // Navigate to page without Widget instantiation
+  await navigate('widget');
+
+  const loginLink = page.getByRole('link', { name: 'Login via Modal Widget' });
+
+  await expect(loginLink).toBeVisible();
+
```
cerebrl, line 14: Added a bit more to this basic test to ensure the misc APIs work before and after logging in.

tests/widget/modal/widget-modal.login.test.js
```diff
@@ -19,4 +26,13 @@ test('Modal widget with login', async ({ page }) => {
...
   await verifyUserInfo(page, expect);
+
+  // Navigate to page without Widget instantiation
+  await navigate('widget');
+
+  // Refresh the page to clear out previous Widget references
+  await page.reload({ waitUntil: 'networkidle' });
+
+  // Test the config and user API without Widget references
+  await verifyUserInfo(page, expect);
```
cerebrl, line 37: Navigate back to the basic page and ensure user information renders without widget instantiation.

README.md
```diff
@@ -396,6 +430,65 @@ NOTE: For more SDK configuration options, please [see our SDK's configuration do
...
+const widget = new Widget({
+  target: document.getElementById('widget-root'),
+  props: {
+    config: {
+      /**
+       * REQUIRED; SDK configuration object
+       */
+      serverConfig: {
+        baseUrl: 'https://customer.forgeblocks.com/am',
+      },
```
daveadams56, line 450: Just something to note, I found that it is mandatory to set a timeout in the `serverConfig`. The examples here imply that it is optional, however an error is thrown if you follow this:

```
ERROR Error: Uncaught (in promise): [
  {
    "code": "invalid_type",
    "expected": "number",
    "received": "undefined",
    "path": [
      "serverConfig",
      "timeout"
    ],
    "message": "Required"
  }
]
```

I can see in [sdk.config.ts](https://github.com/ForgeRock/forgerock-web-login-framework/blob/85de439122efc9b80cf432846fba716c3037631e/src/lib/sdk.config.ts#L52) that there is a comment acknowledging that it should be optional. Given this, I have two questions:

1) Do we have a ticket open for this on the JS SDK? I couldn't see one but maybe I was searching for the wrong thing.
2) Should we make this mandatory in the docs until it is fixed in the JS SDK?

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -0,0 +1,253 @@
...
+      // TODO: Move to a shared utility
+      configure({
+        // Set some basics by default
+        ...{
+          // TODO: Could this be a default OAuth client provided by Platform UI OOTB?
+          clientId: 'WebLoginWidgetClient',
+          // TODO: If a realmPath is not provided, should we call the realm endpoint and detect a likely default?
+          // https://backstage.forgerock.com/docs/am/7/setup-guide/sec-rest-realm-rest.html#rest-api-list-realm
+          realmPath: 'alpha',
+          // TODO: Once we move to SSR, this default should be more intelligent
```
daveadams56, line 45: What's SSR?

daveadams56 (approved):
This looks good to me. I mainly focused on using the widget to take advantage of this feature and after a few bumps which we discussed, I managed to separate the widget instantiation from the configuration quite easily - including using the examples in the docs. Just a few points / questions, none of them blocking. Approved!

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -0,0 +1,247 @@
...
+
+export function widgetApiFactory(modal?: Modal) {
+  let journeyStore: JourneyStore;
+  let oauthStore: OAuthStore;
+  let userStore: UserStore;
+
+  let returnError: (response: Response) => void;
+  let returnResponse: (response: Response) => void;
+
+  const configuration = {
```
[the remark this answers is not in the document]
cerebrl: I think we have one more story we need to do to wrap up our widget APIs, so let's add this naming concern to that story.

README.md
cerebrl: I think this is part of the SDK type fixes that we have a story for addressing. @ryanbas21 can you double check this?

src/lib/widget/_utilities/api.utilities.ts
cerebrl: I know we addressed this in our meeting, but just to respond here: SSR is Server Side Rendering.

src/lib/widget/_utilities/api.utilities.ts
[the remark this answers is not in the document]
cerebrl: Jira for finalizing the naming: https://bugster.forgerock.org/jira/browse/SDKS-2367

PR #6, opened 2023-02-02
ci: update-preview-and-whitesource

ryanbas21:
two fixes: One - we need preview -close to have the docker image i made so that is why it fails.

two: we need to add the base branch for whitesource per bruno


daveadams56 (approved):
Looks good to me.

PR #44, opened 2023-02-16
ci(ci.yml): refactor ci for release

ryanbas21:
ci refactor includes adding chromatic back to the pipeline, and running accessibility tests seperately. i refactored to cache npm install and split it into two jobs so that we are not redoing the same steps overr and over

https://bugster.forgerock.org/jira/browse/SDKS-2256

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,7 +63,7 @@ jobs:
           npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
             "npx http-server storybook-static --port 6006 --silent" \
             "npx wait-on tcp:6006 && npm run test:storybook"
-  visual:
+  chromatic:
     runs-on: ubuntu-latest
     if: github.event_name != 'pull_request'
     needs: install
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,8 @@
 name: 'CI Pipeline'
 on:
-  pull_request:
+  push:
 jobs:
-  pr:
+  install:
     strategy:
       matrix:
         node: [16.x, 18.x]
@@ -17,16 +17,89 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
           cache: 'npm'
-      - run: npm ci
-      - run: npm run build
-        env:
-          VITE_FR_AM_URL: ${{ secrets.VITE_FR_AM_URL }}
-          VITE_FR_AM_COOKIE_NAME: ${{ secrets.VITE_FR_AM_COOKIE_NAME }}
-          VITE_FR_OAUTH_PUBLIC_CLIENT: ${{ secrets.VITE_FR_OAUTH_PUBLIC_CLIENT }}
-          VITE_FR_REALM_PATH: ${{ secrets.VITE_FR_REALM_PATH }}
-      - name: Publish to Chromatic
-        if: ${{ matrix.os == 'ubuntu-latest' && matrix.node < 17 && !github.event.pull_request.draft }}
-        uses: chromaui/action@v1
+      - uses: actions/cache@v3
+        id: npm-cache
         with:
-          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+          path: |
+            node_modules
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-npm
+      - name: Install dependencies if cache invalid
+        if: steps.npm-cache.outputs.cache-hit != 'true'
+        run: npm ci
+  pr:
+    needs: install
+    strategy:
+      matrix:
+        node: [16.x, 18.x]
+        os: [macos-latest, ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    env:
+      VITE_FR_AM_URL: ${{ secrets.VITE_FR_AM_URL }}
+      VITE_FR_AM_COOKIE_NAME: ${{ secrets.VITE_FR_AM_COOKIE_NAME }}
+      VITE_FR_OAUTH_PUBLIC_CLIENT: ${{ secrets.VITE_FR_OAUTH_PUBLIC_CLIENT }}
+      VITE_FR_REALM_PATH: ${{ secrets.VITE_FR_REALM_PATH }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Restore npm dependencies
+        uses: actions/cache@v3
+        id: npm-cache
+        with:
+          path: |
+            node_modules
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-npm
+      - run: npm run build
+      - name: Build Storybook
+        run: npm run build:storybook --quiet
+      - run: npx playwright install
+      - name: Serve Storybook and run tests
+        if: ${{ runner.os == 'ubuntu-latest' }} # this one hangs on mac os for some reason
+        run: |
+          npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
+            "npx http-server storybook-static --port 6006 --silent" \
+            "npx wait-on tcp:6006 && npm run test:storybook"
+  visual:
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request'
+    needs: install
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          cache: 'npm'
+      - name: Restore npm dependencies
+        uses: actions/cache@v3
+        id: npm-cache
+        with:
+          path: |
+            node_modules
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-npm
+      - name: Install dependencies if cache invalid
+        if: steps.npm-cache.outputs.cache-hit != 'true'
+        run: npm ci
+      - uses: chromaui/action@v1
+        with:
+          onlyChanged: true
+          traceChanged: true
+          diagnostics: true
+          buildScriptName: 'build:storybook'
           exitZeroOnChanges: true
+        env:
+          CHROMATIC_PROJECT_TOKEN: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: chromatic-build-artifacts-${{ github.run_id }}
+          path: |
+            chromatic-diagnostics.json
+            **/build-storybook.log
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -2,4 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 CI=true npm run ci
-npm run test:e2e
+CI=true npm run test:e2e
--- a/package.json
+++ b/package.json
@@ -4,7 +4,10 @@
   "scripts": {
     "build": "npm run build:widget && npm run build:app",
     "build:app": "vite build",
-    "build:widget": "rollup -c",
+    "build:dev:widget": "rollup -c",
+    "build:dev:svelte-package": "svelte-package",
+    "build:dev:types": "rollup -c ./rollup.types.config.js",
+    "build:widget": "npm-run-all build:dev:*",
     "build:storybook": "storybook build",
     "build:release": "npm-run-all build:release:*",
     "build:release:widget": "rollup -c",
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "check:lint": "prettier --ignore-path .gitignore -w --plugin-search-dir=. . && eslint --fix --ignore-path .gitignore .",
     "check:svelte": "svelte-check --tsconfig ./tsconfig.json --threshold=warning",
     "check:watch": "svelte-check --tsconfig ./tsconfig.json --watch",
-    "chromatic": "chromatic --project-token=e10acf0c74f9 --force-rebuild",
+    "chromatic": "chromatic --force-rebuild",
     "ci": "vitest --dir src",
     "ci:e2e": "playwright test",
     "ci:e2e:debug": "PWDEBUG=1 playwright test",
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "chromatic": "^6.17.0",
     "color": "^4.2.3",
     "commitizen": "^4.2.6",
+    "concurrently": "^7.6.0",
     "cz-conventional-changelog": "^3.3.0",
     "eslint": "^8.27.0",
     "eslint-config-prettier": "^8.5.0",
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+npx lint-staged & CI=true npm run test 
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,6 +57,7 @@
         "chromatic": "^6.17.0",
         "color": "^4.2.3",
         "commitizen": "^4.2.6",
+        "concurrently": "^7.6.0",
         "cz-conventional-changelog": "^3.3.0",
         "eslint": "^8.27.0",
         "eslint-config-prettier": "^8.5.0",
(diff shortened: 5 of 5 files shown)


daveadams56 (commented):
Quick first pass, think I understand what's been implemented and no problems with changes so far. Will come back to it later this week or when the ticket gets moved over.

daveadams56 (commented):
Left a question that I think was very similar to one that must now be outdated as I can't find it, it's on the format of the `ci.yml` file. Aside from that, I'll message you on slack to ask more context questions.

.github/workflows/ci.yml
```diff
@@ -1,8 +1,8 @@
 name: 'CI Pipeline'
 on:
-  pull_request:
+  push:
```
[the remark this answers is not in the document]
ryanbas21: Yup! https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows 

its pretty much the same as on: [push]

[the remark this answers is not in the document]
ryanbas21: valid yaml or not unsure but valid actions syntax

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
+      - name: Build Storybook
+        run: npm run build:storybook --quiet
+      - run: npx playwright install
+      - name: Serve Storybook and run tests
+        if: ${{ runner.os == 'ubuntu-latest' }} # this one hangs on mac os for some reason
+        run: |
+          npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
+            "npx http-server storybook-static --port 6006 --silent" \
+            "npx wait-on tcp:6006 && npm run test:storybook"
+  visual:
```
daveadams56, line 66: Can you explain what the `visual` job is for please?

.github/workflows/ci.yml
```diff
@@ -1,8 +1,8 @@
 name: 'CI Pipeline'
```
daveadams56, line 1: In the JIRA I see the question "Do we want a nightly run of our automation?" - what did we decide for this? I don't see anything in this PR so guessing we're not doing it here. Otherwise if we are, it would be good if you could point it out. Thanks.

daveadams56 (approved):
I spent a bit more time looking at this today, reviewed the checks and the flows that are defined and think I now understand the steps we're implementing here. I don't have any blockers, just questions about a few things in the JIRA that I can't seem to see here, and one for my understanding. Overall this looks good and I like the improvements. Happy to approve.

.github/workflows/ci.yml
ryanbas21: It's just running the visual tests that storybook has - the ones that above, hang on mac.

.github/workflows/ci.yml
ryanbas21: I don't see a value in doing so - i'm unsure what it would give us that we don't currently have. I think justin agrees, but if you have an idea of something it would give us that we don't currently have i'm happy to hear it

.husky/pre-push
```diff
@@ -2,4 +2,4 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 CI=true npm run ci
-npm run test:e2e
+CI=true npm run test:e2e
```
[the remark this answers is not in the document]
daveadams56: Oh ok, that explains it then. Thanks.

.github/workflows/ci.yml
daveadams56: Understood, thanks for explaining.

.github/workflows/ci.yml
daveadams56: No I'm ok with not having it, given that we run pretty comprehensive testing with CI. Only if we wanted to do some lengthy extra testing nightly on top of what we do for CI, would I think it's worth it. Happy with your answer.

cerebrl (changes requested):
Here's what I believe is missing in this PR:

- The Chromatic token is still in our package.json (all secrets should be removed)
- I don't see our type generation addressed for local dev (see comment on Jira ticket)
- I can't see Chromatic being address explicitly; I see something called "visual" but unsure what that is
- Addressing the Chromatic build limit until we get the approval for paying for the service (my suggestion would be to limit the run to PRs only with a "ui" label on them)

.husky/pre-push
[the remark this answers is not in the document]
ryanbas21: If you run the following command locally, does it successfully quit?

```
      npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
            "npx http-server storybook-static --port 6006 --silent" \
            "npx wait-on tcp:6006 && npm run test:storybook"
```

This is what storybook does with small changes to use our scripts - and i couldn't get this working on prepush or mac ci as it didn't exit gracefully. That being said - I imagine you ran it locally fine because you had two seperate processes. Did you ever run it with the same type of "single command" and successfully exit? 

I'm not 100% keen on adding a ton of things to "pre-push" myself, I don't even love the e2e's running but if we feel its a super hard requirement i'm happy to keep looking.

I got the command from here: https://storybook.js.org/tutorials/ui-testing-handbook/react/en/automate/

.github/workflows/ci.yml
ryanbas21: I mis-spoke, and edited the above comment to reflect hte correct wording. It is the chromatic test run. 

We can absolutely add the label requirement, i'd probably move it out of this workflow and into its own, if we prefer only running the tests when it has a label.

ryanbas21:
> Here's what I believe is missing in this PR:
> 
> * The Chromatic token is still in our package.json (all secrets should be removed)
  - didnt see it in package.json ive added a comit which removes it. 
> * I don't see our type generation addressed for local dev (see comment on Jira ticket)
  - didn't notice this on the jira so ill look at it now.
> * I can't see Chromatic being address explicitly; I see something called "visual" but unsure what that is
  - https://github.com/ForgeRock/forgerock-web-login-framework/pull/44/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR90 (see other comments that link the storybook docs) 
> * Addressing the Chromatic build limit until we get the approval for paying for the service (my suggestion would be to limit the run to PRs only with a "ui" label on them)
  - We could do this.

.github/workflows/ci.yml
```diff
@@ -1,8 +1,8 @@
...
 on:
-  pull_request:
+  push:
 jobs:
-  pr:
+  install:
```
ryanbas21, line 5: Install job - run a matrix of os's and install then cache all node_modules. This will precede any other run and we can restore this from cache later, that way we dont install node modules for every single job / step that requires the same dependencies.

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
-      - run: npm ci
+      - uses: actions/cache@v3
+        id: npm-cache
+        with:
+          path: |
+            node_modules
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-npm
+      - name: Install dependencies if cache invalid
```
ryanbas21, line 28: If we have stuff in cache that matches this hash - use it. if we do not, run clean install.

clean install runs an install from the package-lock so it should copy the exact node modules that were used to build this PR.

npm install - will install and try to pull more recent versions given the `^` or `~` etc.

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
+        if: steps.npm-cache.outputs.cache-hit != 'true'
+        run: npm ci
+  pr:
+    needs: install
+    strategy:
+      matrix:
+        node: [16.x, 18.x]
+        os: [macos-latest, ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    env:
```
ryanbas21, line 38: Add these variables in the job scope as they are reused a bit.

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
+      VITE_FR_AM_URL: ${{ secrets.VITE_FR_AM_URL }}
+      VITE_FR_AM_COOKIE_NAME: ${{ secrets.VITE_FR_AM_COOKIE_NAME }}
+      VITE_FR_OAUTH_PUBLIC_CLIENT: ${{ secrets.VITE_FR_OAUTH_PUBLIC_CLIENT }}
+      VITE_FR_REALM_PATH: ${{ secrets.VITE_FR_REALM_PATH }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Restore npm dependencies
```
ryanbas21, line 47: Restore node modules from the install step

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
-      - name: Publish to Chromatic
-        if: ${{ matrix.os == 'ubuntu-latest' && matrix.node < 17 && !github.event.pull_request.draft }}
-        uses: chromaui/action@v1
-        with:
-          projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
+      - name: Build Storybook
+        run: npm run build:storybook --quiet
+      - run: npx playwright install
+      - name: Serve Storybook and run tests
+        if: ${{ runner.os == 'ubuntu-latest' }} # this one hangs on mac os for some reason
```
ryanbas21, line 61: because i could only get this to run successfully on Ubuntu thats how i have this here.

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
+    needs: install
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
+          cache: 'npm'
+      - name: Restore npm dependencies
```
ryanbas21, line 78: restore dependencies from install step

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
+        uses: actions/cache@v3
+        id: npm-cache
+        with:
+          path: |
+            node_modules
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-npm
+      - name: Install dependencies if cache invalid
```
ryanbas21, line 87: if its no good - install but it should be good usually.

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
+        if: steps.npm-cache.outputs.cache-hit != 'true'
+        run: npm ci
+      - uses: chromaui/action@v1
```
ryanbas21, line 90: chromatic action - run chromatic tests.

.husky/pre-push
ryanbas21, line 5: added headless instead of headed brwoser on push.

.github/workflows/ci.yml
```diff
@@ -17,16 +17,89 @@ jobs:
...
+      - run: npx playwright install
+      - name: Serve Storybook and run tests
+        if: ${{ runner.os == 'ubuntu-latest' }} # this one hangs on mac os for some reason
+        run: |
+          npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
+            "npx http-server storybook-static --port 6006 --silent" \
+            "npx wait-on tcp:6006 && npm run test:storybook"
+  visual:
+    runs-on: ubuntu-latest
+    if: github.event_name != 'pull_request'
```
ryanbas21, line 68: so chromatic needs to run on push according to the docs. I believe this is because they require a commit histoyr that you get from the `push` event. 

This adds complexity because we can't just trigger it on a label or a draft state because those are pull_request event values (my understanding).

So - the `onlyChanged` argument should help alleviate some builds but - it won't be perfect.

Thoughts @daveadams56 @cerebrl ? I know this isn't ideal. But thank you for reminding me as i forgot this.

package.json
```diff
@@ -4,7 +4,10 @@
   "scripts": {
     "build": "npm run build:widget && npm run build:app",
     "build:app": "vite build",
-    "build:widget": "rollup -c",
+    "build:dev:widget": "rollup -c",
+    "build:dev:svelte-package": "svelte-package",
+    "build:dev:types": "rollup -c ./rollup.types.config.js",
+    "build:widget": "npm-run-all build:dev:*",
```
cerebrl, line 10: I see you've added some additional commands here, but what's your recommendation for adding them into the `npm run dev` command? The issue is this command currently doesn't build any types, so when using the e2e apps for local dev verification, you end up with a bunch of type errors. One then has to run these separate commands to fix the errors for each iteration of develop-test-develop cycle. It's not a huge deal, but it just adds up if making significant changes over time.

It would be nice if the types just built with everything else with the one `npm run dev`. If this is too process intensive, is there a different way you would recommend to reference types from the e2e apps to both avoid errors and not have to run separate commands?

.github/workflows/ci.yml
cerebrl: Does it just break Chromatic entirely? What is the loss of functionality if we don't trigger it per push?

The code in it's current state, what does it do? Does it run Chromatic at every push still, or are we not running it at all ... ? Sorry for all the questions, but I feel a bit lost on this particular portion of our CI.

It's also worth noting that running the Storybook tests are not needed in CI if we run Chromatic. They should be mutually exclusive.

package.json
ryanbas21: `build:widget` gets called by `dev` ... right?

.github/workflows/ci.yml
ryanbas21: We discussed this on slack but here is a sumary:

https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events 

`push` means any push event (unless a push has more than 3 tags I think).

It does run it on every push. Its the recommended way. So i'm not 100% sure on what to do.

1. label -> requires the pull_request event
2. draft mode -> requires the pull_request event
3. push -> required for commit history and builds. Previously we were running on pull_request.

So our summary was lets see how it works on push, and see how builds are going. We should then after look to purchase it. In the event we are not allowed to purchase it, and builds are becoming a problem, we will look into another way of handling this. but as of now this is the documented suggested way.

package.json
ryanbas21: Rereading this i think maybe the npm-run-all command is confusing here - it basically runs all of the above commands (i believe in the order they are listed in the file) so any command that matches this pattern

`build:dev:*` so thats how thats working now

cerebrl: `npm run build:widget` only gets called initially, but there's no "watch" command, so one would have to stop and restart the `npm run dev` command. `npm run dev:widget` would be the command that would be best to handle the generation of types. Please test this locally and ensure this isn't too process intensive though. We don't want to slow the dev command to a crawl.

ryanbas21: so is the ask to build the types on every save? If so - just so we understand what we are doing

We save a file:
we all rollup once for the widget itself, source code, one call to svelte-package to save the types, then another call to rollup to bundle the types, and then the app would need to get reloaded with vite.

Is that what we want? I can't imagine this dev flow being sustainable?

ryanbas21: i have to think about this, i wonder if docker is the best way to develop in this case? really not sure

.husky/pre-push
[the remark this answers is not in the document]
daveadams56: The command didn't work at all for me:

```
david.adams@C02G31Q7MD6V forgerock-web-login-framework % npx concurrently -k -s first -n "SB,TEST" -c "magenta,blue" \
            "npx http-server storybook-static --port 6006 --silent" \
            "npx wait-on tcp:6006 && npm run test:storybook"
[TEST] 
[TEST] > @forgerock/login-framework@0.0.1 test:storybook
[TEST] > test-storybook
[TEST] 
[SB] (node:60671) [DEP0066] DeprecationWarning: OutgoingMessage.prototype._headers is deprecated
[SB] (Use `node --trace-deprecation ...` to show where the warning was created)
[TEST] [test-storybook] It seems that your Storybook instance is not running at: http://localhost:6006. Are you sure it's running?
[TEST] 
[TEST] If you're not running Storybook on the default 6006 port or want to run the tests against any custom URL, you can pass the --url flag like so:
[TEST] 
[TEST] yarn test-storybook --url http://localhost:9009
[TEST] 
[TEST] More info at https://github.com/storybookjs/test-runner#getting-started
[TEST] npx wait-on tcp:6006 && npm run test:storybook exited with code 1
--> Sending SIGTERM to other processes..
[SB] npx http-server storybook-static --port 6006 --silent exited with code 0
```

I thought that this command would start storybook? Then after this error I thought hmm maybe it needs to be running - but then got a EADDRINUSE error.

.github/workflows/ci.yml
daveadams56: Makes sense.

.github/workflows/ci.yml
daveadams56: Do we want to create a separate ticket or task to look into fixing this, outside of this PR?

.github/workflows/ci.yml
daveadams56: I'm ok with the explanation in chromatic docs, and then see if it causes a problem running on every push. Happy with your suggested way forward Ryan.

daveadams56:
I've read the responses since I left comments, and I am happy for this to still be approved from my side (no blockers). Once Justin is happy that his points have been resolved hopefully we can get this merged. Thanks for explaining a lot @ryanbas21 - this PR helped me to understand more about what we're doing in CI.

.github/workflows/ci.yml
cerebrl: I'm okay with this running only on one platform. I don't think there's any value to running it on different OSs when it will use the same browser on each OS: Chrome.

.github/workflows/ci.yml
cerebrl: I'd like to ask for a name change since "visual" is misleading. I consider these "integration tests" or "interaction tests". We could even just name it "chromatic". Just not a fan of "visual".

package.json
cerebrl: Being that this is so resource intensive, I think keeping it out of the "watched" changes is okay.

cerebrl (approved):
I think this is good. I'd like to ask for just a name change on the `visual` task, but other than that, this is good.

PR #52, opened 2023-02-21
fix(widget-css): add css as export

ryanbas21:
add css as export to package.json

### DIFF
--- a/package/package.json
+++ b/package/package.json
@@ -9,16 +9,16 @@
     "./modal": {
       "import": "./modal.js",
       "require": "./modal.cjs"
+    },
+    "./widget.css": {
+      "import": "./widget.css",
+      "require": "./widget.css"
     }
   },
   "typesVersions": {
     "*": {
-      "modal": [
-        "./modal.d.ts"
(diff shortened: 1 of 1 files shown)


ryanbas21:
Your preview environment [pr-52-ryanbas21](https://cloud.okteto.com/#/previews/pr-52-ryanbas21) has been deployed with errors.

package/package.json
```diff
@@ -9,16 +9,16 @@
...
   "typesVersions": {
     "*": {
-      "modal": [
-        "./modal.d.ts"
-      ],
-      "inline": [
-        "./inline.d.ts"
-      ]
+      "modal": ["./modal.d.ts"],
+      "inline": ["./inline.d.ts"]
```
cerebrl, line 21: Doesn't have to be included in this PR at all, but can we narrow the the auto-formatting for files only in the `/src` directory. There are many times where these files, often managed by other tools, keep getting flipped back and forth between slightly different formats.

cerebrl (approved):

PR #55, opened 2023-02-23
test: improve test coverage across unit, integration and e2e

daveadams56:
Adding tests for [SDKS-2012](https://bugster.forgerock.org/jira/browse/SDKS-2012).

The tests committed here should all pass if you run them locally - just be mindful that they need to point to Justin's tenant and the journey used by the e2e test is a work-in-progress / likely to change tomorrow when I extend it.


ryanbas21:
Your preview environment [pr-55-ryanbas21](https://cloud.okteto.com/#/previews/pr-55-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-55-ryanbas21.cloud.okteto.net)

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -3,6 +3,7 @@
   import { page } from '$app/stores';
 
   import Widget, { configuration, modal, journey, user } from '$package/modal';
+  import type { Response } from '$lib/widget/interfaces';
```
daveadams56, line 6: Another file with svelte-check "fixes" in it that I needed to add in order to commit.

src/lib/_utilities/style.utilities.ts
```diff
@@ -1,8 +0,0 @@
-export function generateStyleString(styles: Record<string, string>[]) {
```
daveadams56, line 1: As far as I can tell, nothing was using this so I decided to remove it.

.hars/response_was_not_application-json.har
```diff
@@ -1,384 +0,0 @@
-{
```
daveadams56, line 1: I removed this since I'm not sure if we want what I assume is a HAR of an error we saw once, in the repo. Additionally I think I've solved the mystery - if it was a mystery! I got the same error when submitting to the registration tree but not filling in the KBA answer. It seems that the tree fails in that scenario.

src/lib/journey/callbacks/_utilities/callback.utilities.test.ts
```diff
@@ -618,11 +619,11 @@ describe('Username Validation Failure message getter', () => {
         },
       ],
     });
-    const errorMessage = getPasswordValidationFailureText(
+    const errorMessage = getUsernameValidationFailureText(
```
daveadams56, line 622: This is the test I refer to in my comment above, where we've mixed up `getPasswordValidationFailureText` and `getUsernameValidationFailureText`.

src/lib/journey/stages/_utilities/step.utilities.test.ts
```diff
@@ -38,6 +42,26 @@ describe('Test string to key conversion', () => {
   });
 });
 
+describe('Test check validation', () => {
+  it('should return true with failed policies', () => {
+    const checkValidation = initCheckValidation();
+    const result = checkValidation(
```
daveadams56, line 48: If there's a better / more correct way to call the function returned by `initCheckValidation()` I'm open to suggestions.

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
```diff
@@ -21,15 +22,21 @@ test('Modal widget with simple login and misc callbacks', async ({ page }) => {
...
-  // Choice
+  // Choice (select)
   const selectEl = page.getByLabel('Are you sure?');
   await selectEl.selectOption('0');
 
-  // Polling Wait
+  // Choice (radio)
+  const radioEl = page.getByLabel('Are you sure?');
+  await radioEl.selectOption('0');
+  await clickButton('Next', '/authenticate');
```
daveadams56, line 32: Expanded the misc-callbacks test (and corresponding tree) to include choice as radio as well as select, now that we have a way to trigger it through the stage parameter.

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
```diff
@@ -21,15 +22,21 @@ test('Modal widget with simple login and misc callbacks', async ({ page }) => {
...
+
   await Promise.all([
     // NOTE: Make sure timer is same or more than set in Polling Wait node
     page.waitForTimeout(3000),
     page.locator('button', { hasText: 'Next' }).click(),
+    // Spinner
+    expect(page.getByRole('status')).toBeVisible(),
```
daveadams56, line 39: This was the best way I could think of to check that the spinner was visible, using the ARIA status role.

daveadams56:
I have just added a status of the test cases delivered on the testing spreadsheet plus a mention of the ones that were not on there but were added or fixed anyway. A few more to come next week.

tests/widget/modal/widget-modal.api.test.js
```diff
@@ -0,0 +1,27 @@
...
+  await expect(page.getByRole('dialog')).toBeVisible();
+
+  const msgPromise = page.waitForEvent('console');
+
+  // Modal close()
+  await page.getByRole('button', { name: 'Close' }).click();
+
+  // Modal onClose()
+  const msg = await msgPromise;
+  await expect(msg.text()).toEqual('Modal closed due to user');
```
daveadams56, line 24: This method of using console output to verify that something happens when the widget API is called (e.g. `modal.onClose()`), seems flaky after running e2e a few times myself and as part of the pre-push checks. The test worked when I wrote it, but more often than not I have noticed the wrong "next" console message appearing which causes the assertion to fail. I'm going to look into how I can do this in a more robust way, but @cerebrl @ryanbas21 if you have any ideas on how I could test the API events - using console output or not - then I'd be keen to hear.

src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -18,10 +18,10 @@
     userResponse = null;
   }
 
-  form.onMount((component) => console.log(component));
+  form.onMount((component: HTMLDivElement) => console.log(component));
```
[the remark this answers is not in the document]
cerebrl: I consider this a bug in the types for the Widget, and the fix is included in the API refactor PR. Regardless, we can move the fix into whatever PR is going to be used.

src/routes/e2e/widget/modal/+page.svelte
cerebrl: Like the other comment. I consider the need for this a bug. The fix is implemented, but not sure which PR will include it.

src/lib/journey/stages/_utilities/step.utilities.test.ts
cerebrl: Whatever way we can test our code, I'm good with it.

src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -18,10 +18,10 @@
...
   // TODO: Use a more specific type
-  journey.onSuccess((response) => (userResponse = response?.user));
-  journey.onFailure((response) => console.log(response.journey?.error));
+  journey.onSuccess((response: Response) => (userResponse = response?.user));
+  journey.onFailure((response: Response) => console.log(response.journey?.error));
```
cerebrl: Anytime we have to type a callback parameter like this, I would consider it a bug, just FYI.

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
```diff
@@ -2,10 +2,11 @@ import { expect, test } from '@playwright/test';
 
 import { asyncEvents, verifyUserInfo } from '../../utilities/async-events.js';
 
-test('Modal widget with simple login and misc callbacks', async ({ page }) => {
+test.only('Modal widget with simple login and misc callbacks', async ({ page }) => {
```
cerebrl: You have a rouge `.only` here.

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
```diff
@@ -21,15 +22,21 @@ test('Modal widget with simple login and misc callbacks', async ({ page }) => {
...
 
-  // Choice
+  // Choice (select)
   const selectEl = page.getByLabel('Are you sure?');
   await selectEl.selectOption('0');
 
-  // Polling Wait
+  // Choice (radio)
+  const radioEl = page.getByLabel('Are you sure?');
+  await radioEl.selectOption('0');
```
cerebrl, line 31: If we are going to distinguish between these two (select versus radio), I believe this test wouldn't actually catch the difference as they are selected in the same way. The way it is now, this isn't entirely "fail'able". If we narrow the selection by element type, that would help make if fail'able if the component gets the type wrong. I hope that makes sense.

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
cerebrl: This is good. I like it.

src/lib/journey/_utilities/step.mock.ts
```diff
@@ -227,3 +227,20 @@ export const step2 = new FRStep({
...
+  callbacks: [
```
cerebrl, line 234: I know this is a bit silly, but could we add the UsernameCallback to this callback array to align with the intended stage? I'd hate to accidentally misinform a future person joining this project.

cerebrl (changes requested):
The only things that are blockers for me:

- the `.only` in the e2e test
- the select versus radio e2e test
- adding the username callback to the mock UsernamePassword stage

All the comments about type bugs are just to inform. You can leave your changes, and we'll remove them in the same PR that fixes what makes them necessary.

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
daveadams56: Thanks for catching that.

src/routes/e2e/widget/inline/+page.svelte
daveadams56: Understood.

tests/widget/modal/widget-modal.api.test.js
daveadams56: I got some great examples from @cerebrl in the JS SDK project for e2e tests there. This has allowed me to add another bit of test coverage. Thanks!

tests/widget/modal/widget-modal.login-with-misc-callbacks.test.js
daveadams56: This does make sense, however narrowing down by type is trickier than I'd anticipated. The way to do it seems to be by "role" which we do elsewhere in this PR for spinners. Setting it for "select" is fine as that can map to a "menu", however radio buttons (more specifically groups of radio buttons), are trickier. The reason is that there are aria attributes that we do not currently use and expectations of the group of radio inputs being laid out and labelled in a certain way. Some details are under mozilla docs for [radiogroup](https://developer.mozilla.org/en-US/docs/web/accessibility/aria/roles/radiogroup_role). I think I might have to revisit this in a future PR.

cerebrl: Let's just add a TODO comment here to figure out how to differentiate between the two in the future.

cerebrl (approved):
This looks good. Let's get these tests in.

ryanbas21 (approved):

PR #62, opened 2023-02-28
refactor(widget-api): changing widget API over to observables with Svelte stores

cerebrl:
## DO NOT MERGE!

This PR is just to easily reference the changes. We should merge to `alpha` for extended testing before merging to `beta`.

### Unified Widget import!

You no longer have to import from one of the two form factors. They have been unified under one file, the `index.svelte` component. Importing should look like this now:

```js
import Widget, { configuration, component, journey, user } from '@forgerock/login-widget';

// Below defaults to the "modal" form factor
new Widget({ target: widgetEl });
// Or, you can declare the type in the `props`
new Widget({ target: widgetEl, props: { type: 'modal' }});

// For the inline form factor, just use type "inline" in the props
new Widget({ target: widgetEl, props: { type: 'inline' }});
```

### New Supportive APIs

Let me first explain what I dislike about our current API:

1. The journey API was event based (onSuccess, onFailure), kind of like an observable, but it technically wasn’t and was a bit fragile (I was subscribing and unsubscribing from stores, passing  a lot of references around … )
2. The other APIs were async-await
3. async-await doesn’t lend well to granular async control: not initiated versus loading versus success or error
4. Error handling was non-existent in the non-journey APIs
5. I didn’t like the inconsistency between journey and non-journey APIs

As a reminder, here’s an example of our current APIs:

```js
journey.start();
journey.onSuccess((response) => console.log(response));
journey.onFailure((response) => console.log(response));

try {
  const response = await user.authorize();
  console.log(response);
} catch (err) {
  console.log(err);
}

try {
  const response = await user.tokens();
  console.log(response);
} catch (err) {
  console.log(err);
}
```

The new APIs: everything under the covers is a Svelte store, so they are all simple observables. The devs won’t have to import Svelte stores, just FYI. So, all APIs return an event subscription that you then pass a function to be run on each event (a change to the underlying store’s state).

Here’s the same APIs from above using this new design:

```js
const journeyEvents = journey();
journeyEvents.start();
journeyEvents.subscribe((event) => console.log(event));

const authorizeEvents = user.authorize();
authorizedEvents.subscribe((event) => console.log(event))

const tokenEvents = user.tokens()
tokenEvents.subscribe((event) => console.log(event));
```

Each event generating method – `journey`, `user.info`, `user.tokens` – returns a `subscribe` method and `get` or `start` method. These are observables (aka Svelte Stores), so `subscribe` takes a function that is called on each event, so it's intended to be long lived. These are great as top level components that are preserved, or central state management modules the app uses.

*_Note: Upon subscription, they will return the local state_*.

The second method, `get` or `start`, are methods that fetch new data from the server. They will also update the underlying stores leading to an event emission which will call the `subscribe` method. They can be called in two ways:

1. Call and forget: `userInfoEvent.get()`. This will make a server query to AM and update the store when complete, triggering subscribe events.
2. Call with `await` and receive the state only upon completion: `const userInfoCompleteEvent = await userEvent.get();`

Method #1 is recommended as it embraces observables and is a touch more efficient. #2 is an option if you are within a portion of your app where you're only interested in the single, completed state of your request. Or, within a component that is short lived.

To illustrate this, let's use getting tokens as an example:

```js
// Get current, local tokens from the Widget store
const tokenEvents = user.tokens();
tokenEvents.subscribe((event) => console.log(event));
```

```js
// Get remote tokens from server, but use observable
const tokenEvents = user.tokens();
tokenEvents.get();
tokenEvents.subscribe((event) => console.log(event)); // Will fire on each event, as long as subscribed
```

```js
// Get remote tokens from server, but use promise
const tokenEvents = user.tokens();
const tokenCompleteEvent = await tokenEvents.get(); // Returns only the final, completion state
```

Lastly, let's look at the new `component` API. This replaces the old `modal` API with the `modal.onMount` and `modal.open`, etc. The inline also had a `form` API that only had an `onMount` method. All of this has been unified into the one `component` API. It has the same design pattern as the other APIs:

```js
const componentEvents = component();
component.subscribe((event) => console.log(event));
component.open();
component.close();
```

The `subscribe` replaces the `onMount` and `onClose`. It can also be placed anywhere, as long as the references to the returned object (aka `componentEvents`) is preserved, in your application. If you subscribe after `Widget` instantiation or before, the `subscribe` will always return a value with the current state of the component, so race conditions shouldn't happen anymore.

What I like about the new APIs:

1. They are now all consistent API method patterns
3. We aren’t “inventing” a new pattern
4. We are not limiting our APIs with async-await or our own home-grown pattern
5. We can now emit many events that the dev can use or not use, these events are reflected in the state properties `event.loading`, `event.completed`, `event.success`, `event.error`
6. All of these events have the same event interface (more consistency)
7. I was able to reduce lines of code and programmatic complexity

The downside is I had to remove some “nice to haves”:

1. The dev has to explicitly call `journey.start` when using the modal
2. The dev can’t arbitrarily place journey.onSuccess; they have get the event emitter (aka `subscribe`) from calling `journey.start`


ryanbas21:
Your preview environment [pr-62-ryanbas21](https://cloud.okteto.com/#/previews/pr-62-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-62-ryanbas21.cloud.okteto.net)

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -1,254 +1,213 @@
...
+    const requestsOauth = options?.oauth || true;
+    const requestsUser = options?.user || true;
+    const { subscribe } = derived(
+      [journeyStore, oauthStore, userStore],
+      ([$journeyStore, $oauthStore, $userStore], set) => {
+        set({
+          journey: $journeyStore,
+          oauth: $oauthStore,
+          user: $userStore,
+        });
```
cerebrl, line 96: Here, we're using a `derived` store to create a single store out of the three stores, `journey`, `oauth`, and `user`. This single store will emit an event when any of its child stores emit an event. We have complete and fine-grain control over what we emit to the developer, from an event standpoint and a data standpoint. It can be as much or as little as we want.

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -1,254 +1,213 @@
...
-            });
-          if (modal) {
-            modal.close({ reason: 'auto' });
+        if ($journeyStore.successful && $oauthStore.successful && $userStore.completed) {
+          modal && modal.close({ reason: 'auto' });
+        } else if ($journeyStore.successful && $oauthStore.successful) {
+          if (requestsUser && $userStore.loading === false && $userStore.completed === false) {
+            userStore.get();
+          } else if (!requestsUser) {
+            modal && modal.close({ reason: 'auto' });
```
cerebrl, line 104: Much of this logic is just a simplified version of what already existed. Using a derived store was much easier because I didn't have to subscribe and unsubscribe to each store as the widget progressed through the auth flow.

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -1,254 +1,213 @@
...
-        if (response.successful) {
-          userStoreUnsub();
-        }
-      });
+      },
+    );
 
-      if (options?.resumeUrl) {
-        journeyStore.resume(options.resumeUrl);
+    function start(startOptions?: JourneyOptionsStart) {
```
cerebrl, line 116: Here we're creating the `start` method and providing a returned Promise if the dev is only interested in the final event and its state. So, a dev can call `start` by itself and not assign the result of function call to anything, capturing the data in the `subscribe`. Or, the dev can use the `async-await` pattern and capture the returned result as a one-off.

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -1,254 +1,213 @@
...
+            }
+          });
+        });
       }
-      return get(userStore).response;
+
+      return { get: wrappedGet, subscribe };
     },
-    logout: async () => {
+    async logout() {
```
cerebrl, line 155: Here's the biggest exception. Since logout is just an action and doesn't return any data or state, I have it as an async function. I _could_ have this as a method returned in one of the other APIs, user or oauth, but it's not really one or the other, but it affects both. So, for now, it's its own thing.

src/lib/widget/inline.svelte
```diff
@@ -20,71 +20,16 @@
...
-  import type { partialConfigSchema } from '$lib/sdk.config';
-  import type { journeyConfigSchema } from '$journey/config.store';
-  import type { partialStringsSchema } from '$lib/locale.store';
-  import type { partialStyleSchema } from '$lib/style.store';
-
-  export let config: z.infer<typeof partialConfigSchema> | undefined = undefined;
-  export let content: z.infer<typeof partialStringsSchema> | undefined = undefined;
-  export let journeys: z.infer<typeof journeyConfigSchema> | undefined = undefined;
-  export let links: z.infer<typeof partialLinksSchema> | undefined = undefined;
-  export let style: z.infer<typeof partialStyleSchema> | undefined = undefined;
```
cerebrl, line 46: This is all the store related stuff pulled out of the Widget and moved into the API utility.

src/lib/widget/inline.svelte
```diff
@@ -20,71 +20,16 @@
...
-   * Variables with _ are the reactive version of the original variable from above.
-   */
-  api.setJourneyStore(initializeJourney(config));
-  api.setOAuthStore(initializeOauth(config));
-  api.setUserStore(initializeUser(config));
-
-  initializeContent(content);
-  initializeJourneys(journeys);
-  initializeLinks(links);
-  initializeStyle(style);
```
cerebrl, line 87: So much repeated code gone \o/

src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -32,31 +28,36 @@
...
-    journey.start({
+    const { start, subscribe } = journey();
+    start({
       journey: journeyParam || authIndexValue || undefined,
       resumeUrl: suspendedIdParam ? location.href : undefined,
     });
+    subscribe((event: any) => {
+      console.log(event);
+      userResponse = event?.user;
+    });
```
cerebrl, line 60: Here's an example implementation of the new APIs. Notice how I configure the Widget, then instantiate the Widget, then grab the journey API, which returns `start` and `subscribe`. This does introduce a bit more surface area to the APIs, but creates much more consistency across all APIs. It's also much more scalable and flexible for future improvements.

src/routes/e2e/widget/+page.svelte
```diff
@@ -12,30 +12,38 @@
...
+      console.log(event);
+      loading = event.loading;
+      userInfo = event.response;
+    });
+
+    // Using Promise method
+    // const { get } = user.info();
+    // const event = await get();
+    // console.log(event);
+    // userInfo = event.response;
```
cerebrl, line 41: Here's an example of what the APIs look like when you are not using the Widget or `journey` API. This is to simulate a route guard or something.

cerebrl (commented):
Just leaving some comments to help focus your time on the big stuff.

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -94,15 +91,14 @@
     </ul>
     <button on:click={logout}>Logout</button>
   {:else}
-    <button
-      on:click={() =>
-        modal.open({
+    <button on:click={() => {
+        journeyEvents?.start({
```
ryanbas21, line 95: So what happens if multiple different "journeyEvents" variables are scattered throughout an app? I assume because its an observable even if i have say 5 files creating their own "subscriptions" throughout a lifecycle of an app, they are all subscribing to the same source of truth so therefore we don't have to worry about them getting out of sync? Does that make sense / is my logic here correct?

src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -18,10 +17,7 @@
     userResponse = null;
   }
 
-  form.onMount((component) => console.log(component));
-  // TODO: Use a more specific type
-  journey.onSuccess((response) => (userResponse = response?.user));
-  journey.onFailure((response) => console.log(response.journey?.error));
+  form.onMount((component: any) => console.log(component));
```
ryanbas21, line 20: is this a temporary console.log?

src/routes/e2e/widget/+page.svelte
```diff
@@ -12,30 +12,38 @@
...
-      userInfo = await user.info(true);
-    } else {
-      // Set fetched tokens to true so the links render to login
-      fetchedTokens = true;
-    }
+    // Using observable method:
+    const { get, subscribe } = user.info();
+    get();
+    subscribe((event: any) => {
+      console.log(event);
```
ryanbas21, line 32: Temporary?

ryanbas21 (commented):
My initial feelings are pretty straightforward. The api makes sense. I think this is a step in the right direction.

I was writing a test app in solid with the widget and this api feels very in line with a reactive type workflow. (Solid and svelte are very similar), so from this perspective this makes sense.

I do question how this would "feel" in a non reactive/observable workflow, as it would be a bit of a model shift. 

Do you feel the consumer of the code doesn't really need to understand reactivity much at all? I don't see they need to however, i do wonder how someone would feel if they had a few variables that are subscribed to journey events and then they were getting unexpected (from a non reactive way of looking at it) results.

My main blocker is the release process is broken and I just don't know why or how to fix it given this PR. Truly have no idea and i'm not super comfortable releasing without types since we have them.

src/lib/widget/inline.svelte
```diff
@@ -16,75 +16,25 @@
   export const journey = api.journey;
   export const request = api.request;
   export const user = api.user;
+
+  export type ConfigurationApi = ReturnType<typeof api.configuration>;
```
ryanbas21, line 20: ❤️

src/lib/widget/modal.svelte
```diff
@@ -25,47 +21,29 @@
...
-      }
+    open() {
       dialogEl.showModal();
     },
   });
 
   export const configuration = api.configuration;
   export const journey = api.journey;
-  export const modal = api.modal as Modal;
+  export const modal = api.modal as ModalApi;
```
ryanbas21, line 31: Why can't we provide this type, where does it get lost that we need to cast it?

ryanbas21: I see now, thinking if its possible to narrow this somehow

src/routes/e2e/widget/modal/+page.svelte
cerebrl: Yes, they would each have their own subscription to the same underlying stores, so their no syncing issue that I can think of. The only thing we would want to warn against is not littering your codebase of these subscriptions without cleaning them up. This would lead to a memory leak. This is why we should encourage devs to place these at higher level components of their app or within their central state management solutions.

One way to avoid this if they really want to do a lot of one-off throughout their app is to use the Promise-based pattern, instead of the Observable-based pattern. The consequence is they just get the single, final value, but the good is we internally subscribe then unsubscribe for them, so no memory leaks.

src/lib/widget/inline.svelte
cerebrl: Thanks to your teaching, I was able to think of this solution.

src/lib/widget/modal.svelte
```diff
@@ -25,47 +21,29 @@
     onMount(fn: (dialog: HTMLDialogElement, form: HTMLFormElement) => void) {
       callMounted = (dialog, form) => fn(dialog, form);
     },
-    open(options?: JourneyOptions): void {
-      // If journey does not have a step, start the journey
-      if (!get(api.getJourneyStore()).step) {
-        journey.start(options);
-      }
+    open() {
```
daveadams56, line 24: The more I think about the fact that modal.open() and journey.start() are no longer linked, the more I think this is beneficial rather than a drawback. There may well be some scenario where the developer wants to open the modal but allow the user to select something within that modal, before starting the journey - e.g. something that might guide which journey actually gets launched. I'm thinking of when we developers are able to drop their own code into the widget. I think this isn't such a bad thing, even if we factor in the extra line of code to start the journey.

src/routes/e2e/widget/+page.svelte
```diff
@@ -12,30 +12,38 @@
...
+      loading = event.loading;
+      userInfo = event.response;
+    });
+
+    // Using Promise method
+    // const { get } = user.info();
+    // const event = await get();
+    // console.log(event);
+    // userInfo = event.response;
```
daveadams56, line 41: Are we keeping these comments in? Guessing so as it's showing an alternative way to do something but thought I'd check.

src/routes/e2e/widget/modal/+page.svelte
daveadams56: I think this needs to be clearly signposted - are we going to update readme / docs in this PR with warnings or advice, or in a separate one?

daveadams56 (commented):
I really like the way this is going, most of all in consistency and predictability when working with the events. I spent a bit of time reading through your explanation (which by the way makes this a joy to review, spoiled with detail - thanks!). I also spent some time digesting the code changes. I have a few comments on my first thoughts, and I would like to dedicate more time tomorrow to pulling this down and testing in another app to really get to know it. I'll post my findings here. Overall though my feeling is very positive, it feels like the project is maturing / solidifying with this PR. Awesome work!

cerebrl:
@ryanbas21: responding to your summary comment ...

> My initial feelings are pretty straightforward. The api makes sense. I think this is a step in the right direction.

Cool, glad to hear it.

> I was writing a test app in solid with the widget and this api feels very in line with a reactive type workflow. (Solid and svelte are very similar), so from this perspective this makes sense.

Yeah, I think in Solid, they are called "signals", but yes, I really like this pattern as it makes for much more robust UI development.

> I do question how this would "feel" in a non reactive/observable workflow, as it would be a bit of a model shift.

I feel the same way, and don't want to "force" developers into this event-based modal. That's why I provided a Promise-based alternative if they want to avoid observables entirely, or they want just a one-off for some reason.

> Do you feel the consumer of the code doesn't really need to understand reactivity much at all? I don't see they need to however, i do wonder how someone would feel if they had a few variables that are subscribed to journey events and then they were getting unexpected (from a non reactive way of looking at it) results.

I don't think they need to understand the concept deeply, but I think it will be important for us to, at the very least, communicate what the `subscribe` means and provide a link to a good resource for further learning. The devs are just interfacing with the "tailend" of the flow (subscribe) and are not needing to transform the "stream" or pass it on to other observables. So, the surface area to learn is quite small.

The only surprise may come from the fact that when they execute an action, say `user.logout()`, if they subscribe to the `user.info` store, it will automatically update that store and emit events, clearing the user info from the state without them having to do anything. This may be a surprise, but I think a positive one.

> My main blocker is the release process is broken and I just don't know why or how to fix it given this PR. Truly have no idea and i'm not super comfortable releasing without types since we have them.

This should be fixed with the changes I made to the types in the `widget/api.utilities.ts` and `widget/interfaces.ts` files. I see the CI is now green, so we should be good.

src/routes/e2e/widget/+page.svelte
cerebrl: I'm a bit more liberal with what is left behind in these e2e test apps/pages. I'm just illustrating that either are acceptable. In the near future, it might be best to have a switch that gives us a control the flip between the two variations.

src/routes/e2e/widget/modal/+page.svelte
cerebrl: @daveadams56 Absolutely! We will have to rework our documentation to repeatedly ensure we lead them to the right decisions. Documentation will have to be a quick follow up to this story/PR.

ryanbas21 (approved):

daveadams56 (commented):
I built and imported the package into the Angular sample, most of the conversions where I'm using the widget make sense to me, just a few that I'm facing errors on and can't figure out what I'm doing wrong. I've left details in slack. Will pick this up again tomorrow and test further. Thanks.

daveadams56 (approved):
I managed to move past the errors I saw last night thanks to your help on slack @cerebrl - so thank you. I was quickly able to convert all of the usage of the widget over to the new API and tried a mix of subscribing to events, and async / await where it felt like it made sense in my Angular app. This now feels nice to work with! I see no blockers from my side, and would be keen to try this again once we've merged to alpha but overall I am happy to approve.

.scripts/processTypes.mjs
```diff
@@ -18,25 +18,16 @@ const declareModuleSection = `declare module '*.svelte' {
 `;
 const importMetaPath = fileURLToPath(import.meta.url);
 const importMetaDirectoryPath = dirname(importMetaPath);
-const modal = join(importMetaDirectoryPath, '../', 'package', 'modal.d.ts');
-const inline = join(importMetaDirectoryPath, '../', 'package', 'inline.d.ts');
+const index = join(importMetaDirectoryPath, '../', 'package', 'index.d.ts');
```
cerebrl, line 21: Adjusting to our unified component.

package/package.json
```diff
@@ -2,23 +2,13 @@
...
-    "inline": {
-      "import": "./inline.js",
-      "require": "./inline.cjs"
-    },
-    "modal": {
-      "import": "./modal.js",
-      "require": "./modal.cjs"
-    }
+    "import": "./index.js",
+    "require": "./index.cjs"
```
cerebrl, line 6: Since we now have a single index file, I'm assuming we can remove the intermediate object with an "index" key.

package/package.json
```diff
@@ -2,23 +2,13 @@
...
+  "types": "./index.d.ts",
   "typesVersions": {
     "*": {
-      "modal": [
-        "./modal.d.ts"
-      ],
-      "inline": [
-        "./inline.d.ts"
-      ]
+      "index": ["./index.d.ts"]
```
cerebrl, line 11: I'm questioning if we need this `typesVersions`. This seems to be for supporting versions of TypeScript, which I don't think we need to do. If we use the `types` key, I think that's all we need.

src/lib/_utilities/i18n.utilities.ts
```diff
@@ -2,7 +2,7 @@ import sanitize from 'xss';
 import { get } from 'svelte/store';
 import { z } from 'zod';
 
-import { strings } from '$lib/locale.store';
+import { stringsStore } from '$lib/locale.store';
```
cerebrl, line 5: This is to help be more explicit throughout the code. I've had to adjust what's exported by most of these "stores", and changing the name to make it more quickly identifiable.

src/lib/components/compositions/dialog/dialog.svelte
```diff
@@ -1,27 +1,30 @@
...
+  // TODO: Add a keyboard listener
+  export function closeDialog(reason: ComponentStoreValue['reason']) {
     function completeClose() {
       dialogEl?.close();
       dialogEl?.classList.remove('tw_dialog-closing');
 
-      // Call dev provided callback function for event hook
-      closeCallback && closeCallback({ reason });
+      // Ensure we have a store and it has an update method on it
+      componentStore?.update((state) => {
```
cerebrl, line 21: We're directly updating the store here, rather than using the `componentApi`, which as a `close` method, but calling it from here would create a loop in the store, so we directly update the store.

src/lib/journey/config.store.ts
```diff
@@ -43,27 +43,25 @@ const defaultJourneys = {
 // Ensure default follows schema
 journeyConfigSchema.parse(defaultJourneys);
 
-export let configuredJourneys: Readable<StoreItem[]>;
+export let configuredJourneysStore: Writable<StoreItem[]> = writable(
+  (Object.keys(defaultJourneys) as JourneyKeys[]).map((key) => ({
+    ...defaultJourneys[key],
+    key,
+  })),
+);
```
cerebrl, line 51: You'll see this throughout this PR. I've pulled the store creation out of the `initialize` function. I initially wanted to hide the store and only expose the stores in a narrowed way, but I discovered that it limited the flexibility of the framework too much. Having the store exported like this means references don't have to be passed around so much.

src/lib/journey/config.store.ts
```diff
@@ -43,27 +43,25 @@ const defaultJourneys = {
...
 
 export function initialize(customJourneys?: z.infer<typeof journeyConfigSchema> | null) {
   if (customJourneys) {
     // Provide developer feedback if customized
     journeyConfigSchema.parse(customJourneys);
 
     const arr = Object.keys(customJourneys) as JourneyKeys[];
-    configuredJourneys = readable(
+    configuredJourneysStore = writable(
```
cerebrl, line 59: Just saw this. This should be converted into a `set` on the store, rather than recreating the store.

src/lib/widget/_utilities/component.utilities.ts
```diff
@@ -0,0 +1,88 @@
...
+  modal: null,
+  form: null,
+  error: null,
+  mounted: false,
+  open: null,
+  reason: null,
+  type: null,
+});
+
+export const componentApi = () => {
```
cerebrl, line 27: This replaces the `modal` API from the form factor specific Svelte components. Plus, since both modal and inline share the same `mount` method, I could combine the two APIs.

src/lib/widget/index.svelte
```diff
@@ -0,0 +1,68 @@
...
+{:else}
+  <div class="fr_widget-root">
+    <!-- Default `displayIcon` to `true` if `style.stages.icon` is `undefined` or `null` -->
+    <Journey
+      bind:formEl
+      displayIcon={$styleStore?.stage?.icon ?? true}
+      journeyStore={journeyStore}
+    />
+  </div>
+{/if}
```
cerebrl, line 68: I've very excited that this new, unified, Svelte component that now supports both form factors is less than 70 LoC.

src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -32,28 +40,28 @@
...
         },
+        realmPath: 'alpha',
+      },
+      content,
+      links: {
+        termsAndConditions: 'https://www.forgerock.com/terms',
       },
     });
+
+    new Widget({ target: formEl, props: { type: 'inline' }});
```
cerebrl, line 61: This is now the way the dev specifies the form factor. It defaults to modal if `type` isn't given.

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -2,58 +2,37 @@
   import { onMount } from 'svelte';
   import { page } from '$app/stores';
 
-  import Widget, { configuration, modal, journey, user } from '$package/modal';
+  import Widget, { configuration, component, journey, user } from '$package/index';
+
+  const config = configuration();
+  const componentEvents = component();
+  const journeyEvents = journey();
```
cerebrl, line 9: Initialize all the APIs.

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -2,58 +2,37 @@
...
   }
 
-  // TODO: Investigate why the parameter types are needed here
-  modal.onMount((dialog: HTMLDialogElement, form: HTMLFormElement) => {
-    console.log(dialog);
-    console.log(form);
-
-    // Calling this on mount is not good if using HMR as it results in a ton of requests on change
-    // journey.start();
+  componentEvents.subscribe((event) => {
```
cerebrl, line 23: We can now `subscribe` to the `componentEvents`, or whatever you call it, and it can be done anywhere. If you subscribe after the Widget has been initialized, it will just return the current event's object which will indicate if the modal is already mounted or not.

package/package.json
ryanbas21: in the case of one entry file, we can remove this yes.
this was a hack i used to get the two type definitions to work

ryanbas21:
merging and will have a doc PR and fix for test coming soon

PR #106, opened 2023-03-29
fix(base.svelte): fix focus on password field

ryanbas21:
the password confirmation field should not be of focus, and the required fields should be required when a password is inputted

### DIFF
--- a/src/lib/journey/callbacks/password/base.svelte
+++ b/src/lib/journey/callbacks/password/base.svelte
@@ -37,7 +37,7 @@
   let savedValue = '';
   let textInputLabel: string;
   let type: 'password' | 'text' = 'password';
-  let value: unknown;
+  let value: string;
 
   /**
    * @function confirmInput - ensures the second password input matches the first
@@ -74,7 +74,7 @@
     callbackType = callback.getType();
     key = callback?.payload?.input?.[0].name || `password-${callbackMetadata?.idx}`;
     textInputLabel = callback.getPrompt();
-    value = callback?.getInputValue();
+    value = callback?.getInputValue() as string;
 
     /**
      * `savedValue` represents what the user set after blur (local component state)
@@ -120,8 +120,9 @@
 {#if callbackMetadata?.platform?.confirmPassword}
   <ConfirmInput
     forceValidityFailure={doPasswordsMatch === false}
-    isInvalid={doPasswordsMatch === false}
+    passwordsDoNotMatch={doPasswordsMatch === false}
     {key}
+    isRequired={value.length > 0}
     onChange={confirmInput}
     {resetValue}
     showMessage={doPasswordsMatch === false}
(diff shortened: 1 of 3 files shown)


ryanbas21:
Your preview environment [pr-106-ryanbas21](https://cloud.okteto.com/#/previews/pr-106-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-106-ryanbas21.cloud.okteto.net)

src/lib/journey/callbacks/password/base.svelte
```diff
@@ -15,6 +15,7 @@
   import type { Maybe } from '$lib/interfaces';
   import type { CallbackMetadata } from '$journey/journey.interfaces';
   import type { styleSchema } from '$lib/style.store';
+  import Password from './password.svelte';
```
cerebrl, line 18: I think this import is an accident?

ryanbas21: yup, no idea how that was pushed there

src/lib/journey/callbacks/password/confirm-input.svelte
```diff
@@ -41,9 +43,18 @@
   }
 
   $: {
+    isInvalid =
+      passwordsDoNotMatch || (isRequired && typeof value === 'string' && value.length === 0);
```
cerebrl, line 47: I believe the second portion of this condition isn't necessary, now that I think about it. This shouldn't be the only way an input gets marked as invalid. The `form.svelte` component can also mark inputs as invalid. So, I'd recommend you at least attempt to only using `passwordsDoNotMatch` as the value for `isInvalid` and see if everything still works. In other words, remove the `|| (isRequired && typeof value === 'string' && value.length === 0)` from this line.

src/lib/journey/callbacks/password/confirm-input.svelte
```diff
@@ -41,9 +43,18 @@
...
     if (resetValue) {
-     value = undefined;
-     onChange(value);
+      value = undefined;
+      onChange(value);
+    }
+    if (passwordsDoNotMatch) {
+      message = interpolate('passwordConfirmationError', null, 'Passwords do not match');
+    } else if (isRequired && !value) {
```
cerebrl, line 54: I'd like for you to also try removing the second portion of this condition as well. The `&& !value` is likely not needed as the core input component will not show the message unless the input is explicitly marked as invalid. I'd recommend trying to ensure the underlying components work as expected.

ryanbas21: This is correct, it can be removed.

src/lib/journey/callbacks/password/base.svelte
```diff
@@ -36,8 +36,9 @@
   let resetValue = false;
   let savedValue = '';
   let textInputLabel: string;
+  let doesPasswordHaveLength = false;
```
cerebrl, line 39: I think this variable is no longer used, yeah?

cerebrl (approved):
Other than the unused variable, this looks good.

PR #114, opened 2023-04-05
fix(types.ts): rename the types file where we missed it and fix release

ryanbas21:
release in ci was broken because we missed two places to update the types file


ryanbas21:
Your preview environment [pr-114-ryanbas21](https://cloud.okteto.com/#/previews/pr-114-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-114-ryanbas21.cloud.okteto.net)

PR #117, opened 2023-04-17
fix(files): files in package.json means it only updates whats in there

ryanbas21:
remove files because its not needed and using it makes it so only whats in that array is included


ryanbas21:
Your preview environment [pr-117-ryanbas21](https://cloud.okteto.com/#/previews/pr-117-ryanbas21) has been deployed with errors.

PR #119, opened 2023-04-18
Fix types and css

ryanbas21:
Fix the types and css issue that was reported.


cerebrl (approved):

daveadams56 (approved):
Looks good and fixes the issues I originally saw in the Angular widget sample.

PR #120, opened 2023-04-20
docs: Update README to match product documentation. (SDKS-2432)

forgerock-chris:
Updated README using content from the product docs on https://backstage.forgerock.com/docs


cerebrl (approved):
Fully reviewed from the Stash/Bitbucket PR. Looks good to me.

PR #121, opened 2023-05-11
feat(callbacks): device-profile-node addition

ryanbas21:
add the device profile node to the login widget.

https://bugster.forgerock.org/jira/browse/SDKS-2000


ryanbas21:
Your preview environment [pr-121-ryanbas21](https://cloud.okteto.com/#/previews/pr-121-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-121-ryanbas21.cloud.okteto.net)

.svelte-kit/tsconfig.json
cerebrl, line 1: I really wish this file would stop getting changed stylistically.

vitest.config.ts
```diff
@@ -9,6 +9,9 @@ export default defineConfig({
   },
   plugins: [svelte({ hot: !process.env.VITEST })],
   test: {
+    outputDiffMaxLines: Infinity,
+    outputDiffMaxSize: 1000,
+    outputTruncateLength: Infinity,
```
cerebrl, line 14: Can you elaborate on these changes?

vitest.config.ts.timestamp-1683768174756.mjs
cerebrl, line 1: Is this is file some sort of autogenerated backup or duplicate of the `vitest.config.ts`?

src/lib/journey/stages/step.mock.ts
```diff
@@ -1,5 +1,70 @@
...
+          value:
+            '{"identifier":"2258096159-3136287393-51942348","metadata":{"hardware":{"cpuClass":null,"deviceMemory":8,"hardwareConcurrency":16,"maxTouchPoints":0,"oscpu":null,"display":{"width":3440,"height":1440,"pixelDepth":24,"angle":0}},"browser":{"userAgent":"Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36","appName":"Netscape","appCodeName":"Mozilla","appVersion":"5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36","appMinorVersion":null,"buildID":null,"product":"Gecko","productSub":"20030107","vendor":"Google Inc.","vendorSub":"","browserLanguage":null,"plugins":"internal-pdf-viewer;mhjfbmdgcfjbbpaeojofohoefgiehjai;internal-nacl-plugin;"},"platform":{"language":"en-US","platform":"MacIntel","userLanguage":null,"systemLanguage":null,"deviceName":"Mac (Browser)","fonts":"cursive;monospace;sans-serif;fantasy;Arial;Arial Black;Arial Narrow;Arial Rounded MT Bold;Comic Sans MS;Courier;Courier New;Georgia;Impact;Papyrus;Tahoma;Trebuchet MS;Verdana;","timezone":360}},"location":{"latitude":39.999445,"longitude":-105.0224336}}',
```
cerebrl, line 59: Shouldn't this value property be `''`? This mock response should not have the profile pre-populated in it, like the username and password callback just above.

src/lib/journey/callbacks/device-profile/device-profile.svelte
```diff
@@ -0,0 +1,45 @@
...
+  import type { SelfSubmitFunction, StepMetadata } from '$journey/journey.interfaces';
+  import { FRDevice, type DeviceProfileCallback } from '@forgerock/javascript-sdk';
+  import type { Maybe } from '$lib/interfaces';
+  import type { ProfileConfigOptions } from '@forgerock/javascript-sdk/lib/fr-device/interfaces';
+  import Spinner from '$components/primitives/spinner/spinner.svelte';
+
+  export let spinner = false;
+  export let callback: Maybe<DeviceProfileCallback> = null;
+  export let callbackMetadata: Maybe<Record<string, any>> = null;
+  export let deviceConfig: ProfileConfigOptions = {};
```
cerebrl, line 11: I don't think we'd want any device configuration coming from the step/stage component. If we had any custom configuration, we'd likely want that to be set in a Svelte Store when the Login Widget is configured initially (the whole `config.set()` method). Then, in this component, it would access that store for the config information. But, let's skip that for now. We can enhance this in a subsequent update.

src/lib/journey/callbacks/device-profile/device-profile.svelte
```diff
@@ -0,0 +1,45 @@
+<script lang="ts">
+  import type { SelfSubmitFunction, StepMetadata } from '$journey/journey.interfaces';
+  import { FRDevice, type DeviceProfileCallback } from '@forgerock/javascript-sdk';
+  import type { Maybe } from '$lib/interfaces';
+  import type { ProfileConfigOptions } from '@forgerock/javascript-sdk/lib/fr-device/interfaces';
+  import Spinner from '$components/primitives/spinner/spinner.svelte';
+
+  export let spinner = false;
+  export let callback: Maybe<DeviceProfileCallback> = null;
```
cerebrl, line 9: If this component is called, the `callback` type _is_ `DeviceProfileCallback`, so the `Maybe` is not needed.

src/lib/journey/callbacks/device-profile/device-profile.svelte
```diff
@@ -0,0 +1,45 @@
+<script lang="ts">
+  import type { SelfSubmitFunction, StepMetadata } from '$journey/journey.interfaces';
+  import { FRDevice, type DeviceProfileCallback } from '@forgerock/javascript-sdk';
+  import type { Maybe } from '$lib/interfaces';
+  import type { ProfileConfigOptions } from '@forgerock/javascript-sdk/lib/fr-device/interfaces';
+  import Spinner from '$components/primitives/spinner/spinner.svelte';
+
+  export let spinner = false;
```
cerebrl, line 8: Does this need to be exported? Also, I think this component should either render a spinner, when it's the only callback, and nothing when it's composed with other callbacks. That being said, this is internal state and shouldn't be an exposed component prop.

src/lib/journey/callbacks/device-profile/device-profile.mock.ts
```diff
@@ -1,23 +1,59 @@
...
+        },
+        {
+          name: 'message',
+          value: '',
+        },
+      ],
+      input: [
+        {
+          name: 'IDToken1',
+          value:
```
cerebrl, line 24: Let's remove these values here and the one below as they won't come pre-populated like this.

src/lib/journey/_utilities/metadata.utilities.test.ts
```diff
@@ -171,10 +171,10 @@ describe('Test metadata builder function for step', () => {
       ],
     };
     const result = buildStepMetadata(callbackMetadata, stageJson);
-
-    expect(result).toStrictEqual({
+    expect(result).toEqual({
       derived: {
-        isStepSelfSubmittable: false,
+        // im using this function because the function in memroy is different
```
cerebrl, line 176: Can you rewrite this comment?

cerebrl (changes requested):
Just leaving some comments since I was looking over the code. Can you remove all the files in the `components/compositions` and `components/primitives` directories. I think that's leftover from a misunderstanding of the story.

vitest.config.ts
ryanbas21: I can remove them, was trying to get the output of a ttest failure to show up

src/lib/journey/callbacks/device-profile/device-profile.svelte
ryanbas21: yeah i just thought it would be customizable from the component level also but yes the device profile naturally has it

src/lib/journey/_utilities/metadata.utilities.test.ts
```diff
@@ -171,10 +171,11 @@ describe('Test metadata builder function for step', () => {
       ],
     };
     const result = buildStepMetadata(callbackMetadata, stageJson);
-
-    expect(result).toStrictEqual({
+    expect(result).toEqual({
       derived: {
-        isStepSelfSubmittable: false,
+        // because this is a a function now, i'm using the function defined so
+        // the memory is the same and its picked up as such by toEqual.
```
cerebrl, line 177: I think I know what you mean here, but could you reword and clean up the comment a bit for better clarity?

src/lib/journey/_utilities/metadata.utilities.test.ts
```diff
@@ -171,10 +171,13 @@ describe('Test metadata builder function for step', () => {
...
     const result = buildStepMetadata(callbackMetadata, stageJson);
-
-    expect(result).toStrictEqual({
+    expect(result).toEqual({
       derived: {
-        isStepSelfSubmittable: false,
+        // because this is a a function now, i'm using the function defined so
+        // the memory is the same and its picked up as such by toEqual
+        // rather than creating a new function which would be different
+        // in memory.
```
cerebrl, line 179: If this comment is going to persist in our source code, let's ensure it's written well.

```
/**
 * Unlike the other properties, `isStepSelfSubmittable` is a function,
 * so for `toEqual` to pass, the "expected" object needs to reference
 * the same function as the `result` object.
 *
 * TODO: Modify this to assert against the returned value, rather than
 * relying on function reference.
 */
```

src/lib/journey/callbacks/device-profile/device-profile.stories.js
```diff
@@ -0,0 +1,61 @@
...
+  const msg = canvas.getByText('This is a message');
+  await expect(msg).toBeInTheDocument();
+  await Promise.resolve(() => {
+    console.log('waiting for device profile to be gathered');
+  }, 5000);
+
+  const location = cb.isLocationRequired();
+  const metadata = cb.isMetadataRequired();
+
+  const dvice = await device.getProfile({ location, metadata });
```
cerebrl, line 59: Rather than relying on a misspelling of "device", can we just call name this `profile` or `deviceProfile`?

tests/widget/inline/widget-device-profile.test.js
```diff
@@ -0,0 +1,16 @@
+import { expect, test } from '@playwright/test';
+
+import { asyncEvents } from '../../utilities/async-events.js';
+
+test('Inline widget with login in Canadian French', async ({ page, context }) => {
```
cerebrl, line 5: Looks like the test name needs rewording.

tests/widget/inline/widget-device-profile.test.js
```diff
@@ -0,0 +1,16 @@
...
+  const { clickButton, navigate } = asyncEvents(page);
+  await navigate('widget/inline?journey=DeviceProfileNode');
+  await page.getByRole('textbox', { name: 'Username' }).type('demouser');
+  await page.getByRole('textbox', { name: 'Password' }).type('j56eKtae*1');
+  await clickButton('Next', '/authenticate');
+  await page.waitForRequest(/\/authenticate/, async () => {
+    const text = page.getByText('Collecting Profile');
+    await expect(text).toBeVisible();
+  });
```
cerebrl, line 14: I believe this is redundant as the `clickButton` already accounts for waiting until the request is resolved. Once `clickButton` itself resolves, you can test whatever you expect to be in the next view.

tests/widget/inline/widget-device-profile.test.js
```diff
@@ -0,0 +1,16 @@
...
+  await context.grantPermissions(['geolocation']);
```
cerebrl, line 15: I feel like this test isn't finished. We don't seem to be asserting anything, just granting a permission. Can we write this so we assert something regarding device profile to ensure the profile is being generated?

cerebrl (changes requested):

tests/widget/inline/widget-device-profile.test.js
ryanbas21: hmm idk this was required for me to get it working.

ryanbas21: The device profile text disapeared too quickly.

cerebrl: I've pulled this in locally, and I believe I know what's happening. This block of code is never executed. So, the assertion is never applied, which gives you a false sense of "passing". The reason for this is the `/authenticate` request is already resolved within the `clickButton` method, which is awaited, so the following `waitForRequest` event listener is never triggered. If you write a console message within it, you'll never see it logged.

The reason it doesn't work when it's not wrapped in this `waitForRequest` is because the "Collecting Profile" message is never displayed by the UI, ever. It's not that it disappears to quickly. You can see that the message is missing in your GIF provided in the PR description.

That's due to the fact that the node in the journey uses `message` as the key. Keys have to be locales or languages, so the key would be `en` or `en_US` with the value of `"Collecting Profile"`. Once you change the key to `en` you can then directly assert using the below without wrapping it in a `waitFor*`:

```
await expect(page.getByText('Collecting Profile')).toBeVisible();
```

src/lib/journey/stages/stages.stories.js
cerebrl, line 1: We should probably add a story within this that composes device profile with a standard login form ... Just to ensure it composes well without breaking.

tests/widget/inline/widget-device-profile.test.js
cerebrl, line 1: This looks much better. Thanks!

cerebrl (changes requested):
I think this looks great overall. Great improvements. We do have a failed test, and accidential file commit, and some small changes, if you don't mind.

.vscode/launch.json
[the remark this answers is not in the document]
ryanbas21: no im removing it whenever this stupid thing works

src/lib/journey/stages/stages.stories.js
cerebrl: Have you addressed this comment? I don't see this in the updated code.

cerebrl (approved):
Thank you. Looks great.

cerebrl:
Looks like we have a failure on one of the Interaction tests: https://www.chromatic.com/test?appId=64075ad87b9e160633c33298&id=64810969ad7b6340d675c048

ryanbas21:
> Looks like we have a failure on one of the Interaction tests: https://www.chromatic.com/test?appId=64075ad87b9e160633c33298&id=64810969ad7b6340d675c048

yeah i fixed it and forgot to push, sorry

PR #144, opened 2023-08-15
ci(actions): rebase-beta-after-merge-to-main

ryanbas21:
commonly dependabot will create PR's against
main because this is the  target branch set
in github (displayed publicly)  and its annoying to
have to check if main chaniged between   development
cycless. so this is to try to automatically rebase beta
when we merge to main.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,3 +61,9 @@ jobs:
           VITE_FR_AM_COOKIE_NAME: ${{ secrets.VITE_FR_AM_COOKIE_NAME }}
           VITE_FR_OAUTH_PUBLIC_CLIENT: ${{ secrets.VITE_FR_OAUTH_PUBLIC_CLIENT }}
           VITE_FR_REALM_PATH: ${{ secrets.VITE_FR_REALM_PATH }}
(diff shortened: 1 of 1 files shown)


ryanbas21:
Your preview environment [pr-144-ryanbas21](https://cloud.okteto.com/#/previews/pr-144-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-144-ryanbas21.cloud.okteto.net)

.github/workflows/release.yml
```diff
@@ -61,3 +61,9 @@ jobs:
           VITE_FR_AM_COOKIE_NAME: ${{ secrets.VITE_FR_AM_COOKIE_NAME }}
           VITE_FR_OAUTH_PUBLIC_CLIENT: ${{ secrets.VITE_FR_OAUTH_PUBLIC_CLIENT }}
           VITE_FR_REALM_PATH: ${{ secrets.VITE_FR_REALM_PATH }}
+      - name: rebase beta branch
+        if: github.ref == 'refs/heads/main'
```
ryanbas21, line 65: maybe we should only do this when its a dependabot pr that was merged?

PR #150, opened 2023-08-24
chore: update-dependencies

ryanbas21:
https://bugster.forgerock.org/jira/browse/SDKS-2479

Update dependencies, rebased with latest main because of dependabot updates.


ryanbas21:
Your preview environment [pr-150-ryanbas21](https://cloud.okteto.com/#/previews/pr-150-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-150-ryanbas21.cloud.okteto.net)

.storybook/main.js
```diff
@@ -10,17 +11,13 @@ module.exports = {
...
-          implementation: require('postcss'),
-        },
-      },
-    },
-    'storybook-dark-mode',
     // https://storybook.js.org/addons/storybook-addon-code-editor
     // 'storybook-addon-code-editor',
+    'storybook-dark-mode',
+    {
+      name: '@storybook/addon-styling',
```
cerebrl, line 18: I actually had a personal todo for this, so glad to see it here :)

src/lib/journey/callbacks/kba/kba-create.stories.js
```diff
@@ -38,71 +38,50 @@ Interaction.play = async ({ canvasElement }) => {
...
   await userEvent.type(answerEl, 'Red', {
-    delay: 200,
-  });
-
-  await new Promise((resolve) => setTimeout(resolve, 1000));
-
-  await userEvent.selectOptions(selectEl, '2', { delay: 200 });
+    delay: 100,
+  }),
+    await waitFor(() => userEvent.selectOptions(selectEl, '2', { delay: 100 }));
```
cerebrl, line 59: I was originally using the `delay` and the `setTimeout` to mimic a user actually typing and clicking to actually see the interaction in human time. If we make them really fast and remove the `setTimeout`s in between actions, I doubt anyone's going to be able to observe what's really happening. 

If we do that, let's just remove them all together (both `delay`s and `setTimeout`s). They are not really needed, and I don't think anyone's using them to validate UI behavior in Storybook itself.

Plus, aren't these equivalent?

```
await waitFor(() => userEvent.selectOptions(selectEl, '2', { delay: 100 }));
```
```
await userEvent.selectOptions(selectEl, '2', { delay: 100 });
```

Anyway, I'd just drop it all to this:

```
await userEvent.selectOptions(selectEl, '2');
```

.storybook/test-runner.ts
cerebrl, line 1: Can I get a little info on this file? What's the difference between this and what we used to have with the a11y checks? Is this just a replacement, or is it new? Also, did we intend on leaving that bit of code commented out?

ryanbas21: https://storybook.js.org/docs/react/writing-tests/accessibility-testing#setup 

injects actual "validation" of the a11y tests into our test runner. its "disabled" because it would fail us but in the thread in slack i mentioned i set this up and we can include it in our actual pipeline if we choose.

src/lib/journey/callbacks/kba/kba-create.stories.js
ryanbas21: They aren't equivalent. I used waitFor because of failures. but i can look more into this if we want.

waitFor will retry until the expectation is satisfied. I only used waitFor because this test was failing after upgrading and it was how i got it to work. i saw it used in storybook docs and opted to try it.

.storybook/test-runner.ts
cerebrl: Can we enable it, but have it report warnings for now? Then, we create stories to start chipping away at the failures? Once we fix them, we can then flip it to fail on any issue. Is that possible?

src/lib/journey/callbacks/kba/kba-create.stories.js
cerebrl: I wonder if `selectEl` isn't rendered on the page at that moment. Maybe the unintentional side-effect of the delays gave it time to render, but upon removal of the delays, the element doesn't not exist? I say that because we use the `await userEvent.selectOptions()` in other places without the need for `waitFor`, so something must be off here.

.storybook/test-runner.ts
ryanbas21: I can look into this, havent dug deeply into this but i imagine its configurable.

.storybook/test-runner.ts
```diff
@@ -0,0 +1,33 @@
...
+    await injectAxe(page);
+  },
+  async postRender(page, context) {
+    const storyContext = await getStoryContext(page, context);
+    // Apply story-level a11y rules
+    await configureAxe(page, {
+      rules: storyContext.parameters?.a11y?.config?.rules,
+    });
+
+    // Enable this to start having Accessibility checks in our tests!
```
ryanbas21, line 23: Enable this in a different story when we want to fix all the accessibility errors. there is not a warning type option i could find.

made: https://bugster.forgerock.org/jira/browse/SDKS-2698

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -22,23 +20,29 @@
...
-    if (event.journey.error || event.oauth.error || event.user.error) {
-      console.log('Login failure event fired');
-    }
-  });
+  journeyEvents.subscribe(
+    (event: {
+      user: { successful: any; error: any };
+      journey: { error: any };
+      oauth: { error: any };
+    }) => {
```
cerebrl, line 36: I feel like we shouldn't have to define these types. They should be automatically typed, or, at the very least, we should be able to import them in, yeah?

src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -23,20 +23,26 @@
     userResponse = null;
   }
 
-  componentEvents.subscribe((event) => {
+  componentEvents.subscribe((event: { lastAction: string }) => {
```
cerebrl, line 26: Having to define these feels a bit odd to me.

cerebrl (approved):
Non-blocking comments about types in our E2E test app. Feels weird, but not hugely important at the moment.

src/routes/e2e/widget/modal/+page.svelte
cerebrl: Were you not able to remove them?

vitest.config.ts.timestamp-1694546809186-0babcaf429f36.mjs
cerebrl, line 1: Did you intend to commit this?

cerebrl (approved):
Just a few non-blocking comments.

vitest.config.ts.timestamp-1694546809186-0babcaf429f36.mjs
ryanbas21: fixed

src/lib/components/primitives/form/form.svelte
```diff
@@ -13,7 +16,7 @@
    * @return {undefined}
    */
   // TODO: Using an `any` to give us time to figure out these weird event types that just changed
```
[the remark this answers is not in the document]
ryanbas21: removed it

PR #184, opened 2023-09-27
feat(css-variables): build initial "design tokens" for run-time customization

juanManuel05:


src/routes/(app)/+layout.svelte
```diff
@@ -55,6 +55,103 @@
...
+      --tw-colors-success-light-l: 50%;
+
+      --tw-colors-warning-dark-hs: 233, 9%;
+      --tw-colors-warning-dark-l: 50%;
+      --tw-colors-warning-light-hs: 455, 78%;
+      --tw-colors-warning-light-l: 50%;
+
+      /** font family*/
+      --tw-font-family-sans-sans-1: ;
+      --tw-font-size-sans-sans-2: ;
```
cerebrl, line 135: I don't see these used in this PR, can you elaborate on their use? I also can't map these to design tokens configured within `tokens.cjs`, at least not how they are declared here.

themes/default/boxes.cjs
```diff
@@ -4,15 +4,17 @@ module.exports = function (theme) {
      * Container theme settings
      */
     '.containing-box': {
-      backgroundColor: theme('colors.background.light'),
+      '--containing-box-background-color': 'hsl(var(--tw-colors-background-light-hs),var(--tw-colors-background-light-l))',
```
cerebrl, line 7: This looks like the right approach here syntactically. 

The one critique I'd have is the verbosity of all the local custom properties I see within all of these CSS rules. Custom properties declared here, within these rules, are not global like they are when declared within `:root {}`. So, you don't have to worry about name collisions. 

That means we can keep them very short, which is something I'd recommend for easier reading and management. As of right now, the line lengths are really long and there's a lot of characters per property reducing readability and adding to potential for mistakes.

For example: rather than this:

```js
'.containing-box': {
  '--containing-box-background-color': 'hsl(var(--tw-colors-background-light-hs),var(--tw-colors-background-light-l))',
  backgroundColor: `var(--containing-box-background-color, ${theme('colors.background.light')})`,
  // ...
}
```

I'd recommend this:

```js
'.containing-box': {
  '--bg-color': 'hsl(var(--tw-colors-background-light-hs), var(--tw-colors-background-light-l))',
  backgroundColor: `var(--bg-color, ${theme('colors.background.light')})`,
  // ...
}
```

src/routes/(app)/+layout.svelte
```diff
@@ -55,6 +55,103 @@
       height: 100%;
       min-height: 100%;
       background-color: #f6f8fa;
+
+    }
+
+    :root {
```
cerebrl, line 61: I know this is probably just for testing purposes, but we won't actually commit all these properties to the project. By default, we won't use the custom properties at all. We will just list them in docs or maybe have an example file of their use.

Plus, I think it would be best if you test this with Storybook, rather than the test app here. So, you could copy and paste these into this storybook file: https://github.com/ForgeRock/forgerock-web-login-framework/blob/beta/.storybook/preview-head.html#L11

cerebrl (commented):
I do a closer scan tomorrow, but I figured I'd give you some feedback right now.

src/routes/(app)/+layout.svelte
cerebrl: I do want to capture all of these converted color values, though. That will save some time for us and our customers if they want to just tweak the existing colors.

src/routes/(app)/+layout.svelte
juanManuel05: i declared becasue according to tokens.cjs line 68 it could be an option. Shall i remove it?
image

themes/default/boxes.cjs
juanManuel05: it makes totally sense, thought of giving a better understanding taking into account the scope of each variable. Do you want me to rename all of the variables?

src/routes/(app)/+layout.svelte
juanManuel05: yeah awesome, i didn't know i could use storybooks for it.

themes/default/boxes.cjs
juanManuel05: working on the renaming of the variabels

src/routes/(app)/+layout.svelte
cerebrl: Oh, I see. 

Just so you know, the array passed into `fontFamily.sans` is just an array of font families, but it's treated as a single value in CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/font-family, a comma delimited list of fonts. So, this is best defined as a single custom property that just takes a comma delimited list of values as seen in the CSS spec.

So, the custom property would look something like this:

```css
:root {
  --tw-font-family-sans: 'Open Sans', Calibri, Arial, Helvetica, sans-serif;
}
```

themes/default/boxes.cjs
cerebrl: Yes, please. The current naming is quite verbose and creates quite a bit of noise when reading.

ryanbas21:
Your preview environment [pr-184-ryanbas21](https://cloud.okteto.com/#/previews/pr-184-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-184-ryanbas21.cloud.okteto.net)

cerebrl (changes requested):
I'm seeing a lot of snapshot failures in our Chromatic. Please start addressing these. It's important to remember that this PR's code changes should have no visual affect on any of our UI, so any change is an issue.

To see this Chromatic report, scroll down to the bottom of your PR summary/conversation page, and then scroll down within the CI/Github Actions pane. At the bottom, you should see a "UI Tests" item. Click on that and view the flagged components that have detected changes.

themes/default/compositions.cjs
```diff
@@ -78,27 +81,35 @@ module.exports = (config, theme) => ({
...
     },
   },
   /**
    * Dialog box theme settings
    */
   '.dialog-box': {
-    backgroundColor: theme('colors.background.light'),
-    border: `1px solid ${theme('colors.secondary.dark')}`,
+    '--bg-color': 'hsl(var(--tw-colors-background-light-hs),var(--tw-colors-background-light-l))',
+    '--border': '1px solid hsl(var(--tw-colors-secondary-dark-hs),var(--tw-colors-secondary-dark-l))',
```
cerebrl, line 110: I'm a bit worried about mixing static values with custom properties for some border declarations. I was thinking color would have to be declared separately for fallbacks to work. But, in my testing, it seems to fallback to the default if the color values are not declared, even though they are mixed with other valid values like `1px` and `solid`. Can you make sure this works cross-browser? If it does, then I guess I'm okay with it.

cerebrl (changes requested):
This is looking better. I left a few comments.

Also, when you update the commit, can you write it to mimic the PR title (I edited it)? Using `LAW` in the scope isn't quite informative enough. 

Thanks, Juan!

themes/default/compositions.cjs
juanManuel05: I was worried about the same thing but I tested it a few times and, like you also noticed, it seems to fallback to the default if no color declared.

cerebrl (approved):
Will wait for CI to pass.

PR #189, opened 2023-10-02
fix(css/widget-api): fix base link specificity issue & oauth-user disablement

cerebrl:
- fixes a bug where a CSS selector for links are not targeted enough, leading to unintended styling of non-widget elements
- fix disabling `user` and `oauth` requests through `journey()` initialization for widget API
- consolidate the import of SDK in `api.utilities.ts`
- fix type issues in both e2e test pages: inline and modal

### DIFF
--- a/src/lib/widget/_utilities/api.utilities.ts
+++ b/src/lib/widget/_utilities/api.utilities.ts
@@ -1,5 +1,10 @@
-import { Config, FRUser, SessionManager, type ConfigOptions } from '@forgerock/javascript-sdk';
-import { HttpClient } from '@forgerock/javascript-sdk';
+import {
+  Config,
+  FRUser,
+  HttpClient,
+  SessionManager,
+  type ConfigOptions,
+} from '@forgerock/javascript-sdk';
 import { derived, get, type Readable } from 'svelte/store';
 
 import { logErrorAndThrow } from '$lib/_utilities/errors.utilities';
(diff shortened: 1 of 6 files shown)


ryanbas21:
Your preview environment [pr-189-ryanbas21](https://cloud.okteto.com/#/previews/pr-189-ryanbas21) has been deployed with errors.

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -139,8 +144,8 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
       logErrorAndThrow('missingStores');
     }
 
-    const requestsOauth = options?.oauth || true;
```
ryanbas21, line 142: Was this causing an issue? I'm  curious i would have expected this to still work just fine?

cerebrl: If the customer does not want the widget to request `user` or `oauth`, they would pass `false` to these config properties, which leads to the right-hand operand being assigned. This means the user essentially can never disable these features, so that's why I switched to the nullish coalescing operator.

ryanbas21: ah, sneaky, sneaky. makes sense.

ryanbas21 (approved):

PR #197, opened 2023-10-09
feat(recaptcha): support-recaptcha-v2

ryanbas21:
support recaptcha v2 in login widget. This does not include v3 support. hcaptcha is supported. invisible recaptcha is not.

https://bugster.forgerock.org/jira/browse/SDKS-2003

### DIFF
--- a/src/lib/journey/callbacks/recaptcha/recaptcha.svelte
+++ b/src/lib/journey/callbacks/recaptcha/recaptcha.svelte
@@ -1,0 +1,79 @@
+<script lang="ts">
+  import type { ReCaptchaCallback } from '@forgerock/javascript-sdk';
+  import type { z } from 'zod';
+  import { onMount } from 'svelte';
+
+  import type { SelfSubmitFunction, StepMetadata } from '$journey/journey.interfaces';
+  import type { styleSchema } from '$lib/style.store';
+  import type { Maybe } from '$lib/interfaces';
+
+  export let callback: Maybe<ReCaptchaCallback>;
+  export const selfSubmitFunction: Maybe<SelfSubmitFunction> = null;
+  export const stepMetadata: Maybe<StepMetadata> = null;
+  export const style: z.infer<typeof styleSchema> = {};
+
+  const siteKey = callback?.getSiteKey() ?? '';
+  let recaptchaApi = `${callback?.getOutputByName<string>('captchaApiUri', '') ?? ''}`;
+  /**
+   * AM defaults the class name to g-captcha which is wrong
+   * I dont think we should be manipulating the class here,
+   * but the classname should be g-recaptcha for google
+   */
+  const recaptchaClass: string =
+    callback?.getOutputByName<string>('captchaDivClass', 'h-captcha') ?? 'h-captcha';
+
+  onMount(() => {
+    window.frHandleCaptchaError = function handleCaptchaError() {
+      callback?.setResult('');
+      if (recaptchaClass.startsWith('h')) {
+        window.hcaptcha.render(recaptchaClass, {
+          sitekey: siteKey,
+          callback: 'frHandleCaptcha',
+          'expired-callback': 'frHandleExpiredCallback',
+          'chalexpired-callback': 'frHandleExpiredCallback',
+          'error-callback': 'frHandleErrorCallback',
+        });
+      } else {
+        window.grecaptcha.render(recaptchaClass, {
+          sitekey: siteKey,
+          callback: window.frHandleCaptcha, // grecaptcha uses different types so passing real function here
+          'expired-callback': window.frHandleExpiredCallback,
+        });
+      }
+    };
+    window.frHandleCaptcha = function handleCaptchaToken(token: string) {
+      callback?.setResult(token);
+    };
+    window.frHandleExpiredCallback = function handleExpiredCallback() {
+      callback?.setResult('');
+      if (recaptchaClass.startsWith('h')) {
+        window.hcaptcha.render(recaptchaClass, {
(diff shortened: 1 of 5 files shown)


ryanbas21:
Your preview environment [pr-197-ryanbas21](https://cloud.okteto.com/#/previews/pr-197-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-197-ryanbas21.cloud.okteto.net)

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -0,0 +1,79 @@
...
+  export const style: z.infer<typeof styleSchema> = {};
+
+  const siteKey = callback?.getSiteKey() ?? '';
+  let recaptchaApi = `${callback?.getOutputByName<string>('captchaApiUri', '') ?? ''}`;
+  /**
+   * AM defaults the class name to g-captcha which is wrong
+   * I dont think we should be manipulating the class here,
+   * but the classname should be g-recaptcha for google
+   */
+  const recaptchaClass: string =
```
cerebrl, line 22: By "class" here, I'm assuming we mean CSS class, yeah? And, this is a reference to the element to which the reCaptcha is expected to mount? I see we are using that class on the element template below, so the Captcha will mount to this component's HTML?

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+  export let callback: Maybe<ReCaptchaCallback>;
+  export const selfSubmitFunction: Maybe<SelfSubmitFunction> = null;
+  export const stepMetadata: Maybe<StepMetadata> = null;
+  export const style: z.infer<typeof styleSchema> = {};
+
+  const siteKey = callback?.getSiteKey() ?? '';
+  let recaptchaApi = `${callback?.getOutputByName<string>('captchaApiUri', '') ?? ''}`;
+  /**
+   * AM defaults the class name to g-captcha which is wrong
```
cerebrl, line 18: "wrong" sounds a bit harsh :)

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -0,0 +1,79 @@
...
+
+  onMount(() => {
+    window.frHandleCaptchaError = function handleCaptchaError() {
+      callback?.setResult('');
+      if (recaptchaClass.startsWith('h')) {
+        window.hcaptcha.render(recaptchaClass, {
+          sitekey: siteKey,
+          callback: 'frHandleCaptcha',
+          'expired-callback': 'frHandleExpiredCallback',
+          'chalexpired-callback': 'frHandleExpiredCallback',
```
cerebrl, line 33: Is `chalexpired-callback` the expected name? Seems like a typo, but if it's correct, it's just a weird abbreviation :)

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -0,0 +1,79 @@
...
+   * but the classname should be g-recaptcha for google
+   */
+  const recaptchaClass: string =
+    callback?.getOutputByName<string>('captchaDivClass', 'h-captcha') ?? 'h-captcha';
+
+  onMount(() => {
+    window.frHandleCaptchaError = function handleCaptchaError() {
+      callback?.setResult('');
+      if (recaptchaClass.startsWith('h')) {
+        window.hcaptcha.render(recaptchaClass, {
```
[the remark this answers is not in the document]
ryanbas21: So they are attached by the script tag below, that is correct.

The `onMount` functions only have to be "complete" in terms of definition, before we attempt to solve the recaptcha. They are "hooks" into the recaptcha essentially. If they aren't defined, it would error when it attempts to call the function.

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
ryanbas21: Correct, the class is the css class, and its the class which hcaptcha or grecaptcha is going to attempt to inject its iframe into.

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
ryanbas21: i can make it "incorrect" if that sounds better 👍

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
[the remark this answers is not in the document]
cerebrl: Okay, so you're saying the `onMount` needs to fire before the user interacts with the Captcha?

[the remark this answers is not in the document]
ryanbas21: Correct, if they interact with the captcha before this none of the callbacks (function callbacks) will fire to set the callback values

ryanbas21:
> Is there anything we can test from an automation standpoint? I guess no. This really is nothing but side-effects and there's no UI from our side ... but I feel like I should ask anyway.

it's a fair question. I can try to write an e2e that clicks a box but it will rely on the recaptcha information of a given user (me) to pass.

I don't love the additional outside requirements, but we could feasibly do it as long as playwright can click the box which I think it should? 

It's almost backwards because recaptcha should prevent bots, but an automated test is basically a bot, so i guess we would be testing that the recaptcha can be clicked but not solved? I'm not entirely sure where the value lies.

cerebrl (approved):

PR #213, opened 2023-10-24
fix(message-node): render-when-userinput-cb

ryanbas21:
moved suspendedtext to its own stage as per comments and updated the metadata to include a userinputcb number to base if a next button needs to be renderd


ryanbas21:
Your preview environment [pr-213-ryanbas21](https://cloud.okteto.com/#/previews/pr-213-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-213-ryanbas21.cloud.okteto.net)

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -22,6 +22,8 @@ export function buildCallbackMetadata(
   stageJson?: Record<string, unknown> | null,
 ) {
   const callbackCount: Record<string, number> = {};
+  const shouldForceButton =
+    step.callbacks.length === 1 && step.callbacks[0].getType() === CallbackType.TextOutputCallback;
```
cerebrl, line 26: I feel like this bit of logic is fairly arbitrary. Let's zoom out a bit and ask if our original assumptions are flawed, because, I feel like they are.

Originally, most of this logic was built around the fact that we treated `TextOutputCallback` and `SuspendedTextOutputCallback` as the same callback. It made sense as it was the only one of the two that came up when we were building the 4 OOTB journeys. It represented the suspension of a journey, and therefore didn't have a "next" button.

But, upon re-evaluation, I think that thinking is flawed. First, `SuspendedTextOutputCallback` doesn't have to be treated in the system just like `TextOutputCallback`. That may be our first mistake.

The second flaw might be that rather than the "rule", `SuspendedTextOutputCallback` might be better treated as the "exception". We may end up with many more cases of `TextOutputCallback` that can be treated as a just a way to display text to a user all while preserving the ability to move forward in the journey.

So, here's what I think we should do:

1. Let's ensure we have no logic making `TextOutputCallback` affect the presence or absence of the form's submit button
2. Let's move the mapping of a step that contains `SuspendedTextOutputCallback` to its own stage component
3. We can still use the existing `TextOutput` component to render the `SuspendedTextOutputCallback`, but let's do it from a dedicated stage component
4. The reason this should be a much easier solution is because the `SuspendedTextOutputCallback` is super easy, fool-proof way to detect this special case

I think the one thing we'll need to change with the original code is to add one more `||` condition here: https://github.com/ForgeRock/forgerock-web-login-framework/blob/beta/src/lib/journey/stages/generic.svelte#L129.

This condition should check if there are any callbacks that accept user input. If there are none, then we should render a submit button because that means this is a step that's just for displaying info to the user, nothing more. So, it would look like this:

```
{#if !metadata?.step?.derived.numOfUserInputCBs.length || metadata?.step?.derived.isUserInputOptional || !metadata?.step?.derived.isStepSelfSubmittable()}
  <Button busy={journey?.loading} style="primary" type="submit" width="full">
    <T key="nextButton" />
  </Button>
 {/if}
```

So, the above would do the following:

1. Are there any callbacks that accept user input? No? Then render the button.
2. Are there any callbacks that accept user input but optional? Yes? Then render button.
3. Is this step self submittable? No? Then render button.

Of course, this new logic would lead to the `SuspendedTextOutputCallback` rendering a next button, which is wrong. That's where we send any step containing that callback to a dedicated stage. Boom, done!

src/lib/journey/_utilities/metadata.utilities.ts
```diff
@@ -94,6 +98,7 @@ export function buildStepMetadata<T = unknown>(
       numOfSelfSubmittableCbs: callbackMetadataArray.filter((cb) => !!cb.derived.isSelfSubmitting)
         .length,
       numOfUserInputCbs: numOfUserInputCbs,
+      forceButton: shouldForceButton ?? false,
```
cerebrl, line 101: This is one thing that gave me pause. This seems very UI-centric to be in a pure data analysis portion of this app. Keeping metadata purely about the data is an important aspect to these functions. They should just describe the data, and the UI should then interpret the data to inform how it should render.

cerebrl (changes requested):

src/lib/journey/stages/generic.svelte
```diff
@@ -125,8 +126,21 @@
       }}
     />
   {/each}
-
-  {#if metadata?.step?.derived.isUserInputOptional || !metadata?.step?.derived.isStepSelfSubmittable()}
+  <!--
+    The below condition follows this logic to render the submit button:
```
ryanbas21, line 130: 👍

PR #246, opened 2023-11-22
feat(passwords-config): add config for show password display UI

juanManuel05:


ryanbas21:
Your preview environment [pr-246-ryanbas21](https://cloud.okteto.com/#/previews/pr-246-ryanbas21) has been deployed with errors.

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -111,6 +111,7 @@
       },
       style: {
         labels: 'floating',
+        showPassword:'checkbox',
```
cerebrl, line 114: Unfortunately, this change means all test runs will have the checkbox style. What we need to do is programmatically set this only when we want to test the checkbox option. The best way to do that is to use URL query parameters.

tests/widget/modal/widget-modal.show-password-checkbox.test.js
cerebrl, line 1: Let's change this test to verify all three options: button, checkbox and disabled using URL query parameters.

tests/widget/modal/widget-modal.show-password-checkbox.test.js
```diff
@@ -0,0 +1,36 @@
...
+  await clickButton('Open Login Modal', '/authenticate');
+
+  await expect(page.getByRole('dialog')).toBeVisible();
+
+  await page.getByLabel('Username').fill('user');
+  await page.getByLabel('Password').fill('password1');
+
+  await page.getByRole('checkbox').check();
+  await expect(page.getByRole('checkbox')).toBeFocused();
+  await expect(page.getByLabel('Password')).toHaveValue('password1');
```
cerebrl, line 35: This isn't technically testing the checkbox functionality. The value itself isn't going to change relative to the show or hide password feature. What changes is the `type` attribute of the `input`, so we should probably test that, instead of the value the input is holding.

cerebrl (changes requested):
I'd like a bit of refactoring here, and some improvements on the testing. I'd also like to see a Storybook integration test as well for this show password feature.

src/lib/style.store.ts
```diff
@@ -14,6 +14,7 @@ export const styleSchema = z
   .object({
     checksAndRadios: z.union([z.literal('animated'), z.literal('standard')]).optional(),
     labels: z.union([z.literal('floating').optional(), z.literal('stacked')]).optional(),
+    showPassword: z.union([z.literal('button').optional(), z.literal('checkbox')]).optional(),
```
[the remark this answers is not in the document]
cerebrl: Adding @ryanbas21 for additional feedback.

src/lib/components/compositions/input-floating/floating-label.svelte
```diff
@@ -50,7 +51,12 @@
     {type}
     bind:value
   />
-  <slot name="input-button" />
+  {#if hasRightIcon}
+    <slot name="input-button" />
+  {:else} 
+    <slot name="input-checkbox" />
+  {/if}
```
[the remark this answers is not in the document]
juanManuel05: done

src/lib/style.store.ts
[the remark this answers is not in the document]
juanManuel05: added as a 'none'.

src/routes/e2e/widget/modal/+page.svelte
juanManuel05: done

src/lib/style.store.ts
```diff
@@ -9,11 +9,12 @@ export const logoSchema = z
     width: z.number().optional(),
   })
   .strict();
-
+  
 export const styleSchema = z
   .object({
     checksAndRadios: z.union([z.literal('animated'), z.literal('standard')]).optional(),
     labels: z.union([z.literal('floating').optional(), z.literal('stacked')]).optional(),
+    showPassword: z.union([z.string(), z.null()]).optional(),
```
cerebrl, line 17: Let's make sure the type is as narrow as possible. What I'd recommend is using `z.literal` rather than the generic `z.string`. Also, I don't think it needs `z.null` if it's optional. The fact that it's optional should allow a _nullish_ value.

@ryanbas21 Can you double check this for me?

cerebrl: Also, I'd like to have `'none'` as an option for the customer, in addition to `'button'`, `'checkbox'` and `undefined`. Can we add this as a possible value, so they can essentially disable the show password feature entirely?

@ryanbas21 Thoughts?

src/lib/components/compositions/input-floating/floating-label.svelte
```diff
@@ -21,6 +21,7 @@
   export let type: 'date' | 'email' | 'number' | 'password' | 'phone' | 'text' = 'text';
   export let value = '';
 
+
```
cerebrl, line 24: Not a huge deal, but if you manually added this blank lines, can you remove them. That way, the PR is not misleading regarding files changed.

src/lib/journey/callbacks/password/password.stories.js
```diff
@@ -48,3 +58,22 @@ Interaction.play = async ({ canvasElement }) => {
   await userEvent.click(eye);
   expect(cb.getInputValue()).toBe('password');
 };
+
+export const InteractionWithCheckbox = Template.bind({});
+
+InteractionWithCheckbox.args = { ...PasswordWithCheckbox.args };
+
+InteractionWithCheckbox.play = async ({ canvasElement }) => {
```
cerebrl, line 66: This is great. Thanks! I'd like to add two more tests: 

1. One that ensures `'button'` works
2. One that ensure the new disable option works with `'none'`

tests/widget/modal/widget-modal.show-password-config.test.js
cerebrl, line 1: This is good. Let's just add the final `'none'` option to the test.

cerebrl (changes requested):
I think this looks great. I'd just like to ask if we can add `'none'` as an option to this `showPassword` property, allowing the customer to disable the feature all together. Then, just add an appropriate test for it as well. I've noted this new addition to the Acceptance Criteria in the Jira ticket for posterity.

src/lib/style.store.ts
ryanbas21: I agree. I don't think the code as is makes a ton of sense.


a union of literals would be more fitting if possible.

So we would have 

```
z.union(z.literal('none'), z.literal('button'), z.literal('checkbox'))
```

src/lib/components/compositions/input-floating/floating-label.svelte
juanManuel05: done

src/lib/style.store.ts
juanManuel05: done

tests/widget/modal/widget-modal.show-password-config.test.js
juanManuel05: done

src/lib/journey/callbacks/password/password.stories.js
juanManuel05: the one that ensures button works is the ona called 'base'. It sets the style to 'button'
```
export const Base = {
  args: {
    callback: step.getCallbackOfType(CallbackType.PasswordCallback),
    **style: {
      showPassword: 'button',
    },**
  },
};
```
I will add another one with the option 'none'

cerebrl (approved):

PR #252, opened 2023-11-28
feat(recaptchav3): implement v3 for recaptcha in LAW

ryanbas21:
recaptcha v3 support added to LAW

https://bugster.forgerock.org/jira/browse/SDKS-2811

### DIFF
--- a/src/lib/journey/_utilities/callback-mapper.svelte
+++ b/src/lib/journey/_utilities/callback-mapper.svelte
@@ -60,6 +60,7 @@
   } from '$journey/journey.interfaces';
   import type { styleSchema } from '$lib/style.store';
   import type { Maybe } from '$lib/interfaces';
+  import type { StepOptions } from '@forgerock/javascript-sdk/src/auth/interfaces';
 
   type Props = {
     callback: FRCallback;
@@ -71,7 +72,11 @@
   export let props:
     | Props
     | (Props & { recoveryCodes: Array<string> })
-    | (Props & { webAuthnValue: WebAuthnStepType });
+    | (Props & { webAuthnValue: WebAuthnStepType })
+    | (Props & {
+        journey: Maybe<StepOptions>;
+        recaptchaAction: string;
+      });
 
   let cbType: string;
 
--- a/src/lib/journey/_utilities/callback-mapper.svelte
+++ b/src/lib/journey/_utilities/callback-mapper.svelte
@@ -60,7 +60,6 @@
   } from '$journey/journey.interfaces';
   import type { styleSchema } from '$lib/style.store';
   import type { Maybe } from '$lib/interfaces';
-  import type { StepOptions } from '@forgerock/javascript-sdk/src/auth/interfaces';
 
   type Props = {
     callback: FRCallback;
@@ -72,11 +71,7 @@
   export let props:
     | Props
     | (Props & { recoveryCodes: Array<string> })
-    | (Props & { webAuthnValue: WebAuthnStepType })
-    | (Props & {
-        journey: Maybe<StepOptions>;
-        recaptchaAction: string;
-      });
+    | (Props & { webAuthnValue: WebAuthnStepType });
 
   let cbType: string;
 
--- a/src/lib/journey/callbacks/recaptcha/recaptcha.svelte
+++ b/src/lib/journey/callbacks/recaptcha/recaptcha.svelte
@@ -6,14 +6,21 @@
   import type { SelfSubmitFunction, StepMetadata } from '$journey/journey.interfaces';
   import type { styleSchema } from '$lib/style.store';
   import type { Maybe } from '$lib/interfaces';
+  import type { StepOptions } from '@forgerock/javascript-sdk/src/auth/interfaces';
 
   export let callback: Maybe<ReCaptchaCallback>;
+  export let journey: Maybe<StepOptions> = null;
   export const selfSubmitFunction: Maybe<SelfSubmitFunction> = null;
   export const stepMetadata: Maybe<StepMetadata> = null;
   export const style: z.infer<typeof styleSchema> = {};
+  // This is the action that the user can pass into the component
+  // if it isn't passed we default to the journey.tree value
+  // journey.tree won't necessarily be set on mount, we have to wait for it
+  // to be passed down so defaulting to an empty string so its falsey
+  export let recaptchaAction = journey?.tree ?? '';
 
   const siteKey = callback?.getSiteKey() ?? '';
-  let recaptchaApi = `${callback?.getOutputByName<string>('captchaApiUri', '') ?? ''}`;
+  let isV3 = callback?.getOutputByName('reCaptchaV3', false);
   /**
    * AM defaults the class name to g-captcha which is wrong
    * I dont think we should be manipulating the class here,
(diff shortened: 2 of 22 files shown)


ryanbas21:
Your preview environment [pr-252-ryanbas21](https://cloud.okteto.com/#/previews/pr-252-ryanbas21) has been deployed with errors.

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -6,14 +6,21 @@
...
   export let callback: Maybe<ReCaptchaCallback>;
+  export let journey: Maybe<StepOptions> = null;
   export const selfSubmitFunction: Maybe<SelfSubmitFunction> = null;
   export const stepMetadata: Maybe<StepMetadata> = null;
   export const style: z.infer<typeof styleSchema> = {};
+  // This is the action that the user can pass into the component
+  // if it isn't passed we default to the journey.tree value
+  // journey.tree won't necessarily be set on mount, we have to wait for it
+  // to be passed down so defaulting to an empty string so its falsey
+  export let recaptchaAction = journey?.tree ?? '';
```
cerebrl, line 20: I seem to be unable to find where `recaptchaAction` is passed into this component. I was expecting to see it as a `journeyEvent.start()` config option and passed through to this component. Am I just missing something?

Also, we'll need to think about where this value custom action value gets stored. Maybe in the journey stack?

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -23,57 +30,79 @@
...
+    }
+  }
+  $: {
+    // if recaptcha Action is 0 length, then it has not been passed in as a prop
+    // and we should use the journey tree value which we may have to wait for since
+    // it is not going to be available on mount immediatly due to be an async function
+    recaptchaAction = recaptchaAction.length === 0 ? journey?.tree ?? '' : recaptchaAction;
+    if (recaptchaAction.length) {
+      executeCaptcha();
+    }
```
cerebrl, line 94: I think some of this can be cleaned up by directly using the actual journey stack store. We can then `subscribe` to the store and operate on the value it emits, which should be the whole stack. Then, you can iterate on that stack to find the journey name value. You could even make a derived store out of that store to give you the reactivity you probably want.

src/lib/journey/stages/generic.svelte
```diff
@@ -119,6 +122,7 @@
       props={{
         callback,
         callbackMetadata: metadata?.callbacks[idx],
+        journey: currentJourney,
```
cerebrl, line 125: If we use the `journey` stack store directly in the callback component, then you won't have to worry about passing this value through the layers, cleaning up the code a bit.

cerebrl (changes requested):
We should probably think about testing as well. Is there any way to appropriately test this? I'm not sure, tbh, as so much of it is all external APIs and side-effect. But, just figured I'd ask.

src/lib/journey/_utilities/callback-mapper.svelte
```diff
@@ -71,7 +72,11 @@
   export let props:
     | Props
     | (Props & { recoveryCodes: Array<string> })
-    | (Props & { webAuthnValue: WebAuthnStepType });
+    | (Props & { webAuthnValue: WebAuthnStepType })
+    | (Props & {
+        journey: Maybe<StepOptions>;
+        recaptchaAction: string;
+      });
```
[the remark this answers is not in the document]
ryanbas21: I think we could probably leverage a discriminated union type but we would be making this type larger. However i think it would improve the intellisense through the switch case.

I'm going to pose this question to some groups and see if i'm on the right track

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
ryanbas21: i must have not committed this...weird.

src/lib/journey/journey.store.ts
cerebrl, line 1: I think the repetition visible here means we might want to refactor this a bit in the future.

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -23,57 +38,59 @@
...
+      renderCaptcha({
+        nameOfCaptcha: recaptchaClass === 'g-recaptcha' ? 'grecaptcha' : 'hcaptcha',
+        siteKey,
+      });
+    }
+  });
+  // defining this outside of the reactive block and guarding it with a isV3 check so it only runs when v3
+  // is defined as true and we have a recaptcha action to assign.
+  function executeV3Captcha() {
+    if (isV3 && recaptchaAction.length) {
```
cerebrl, line 61: Is an action value a required attribute for v3? We provide a default value for `recaptchaAction`, yes? If so, would this ever be an empty string?

src/lib/journey/journey.store.ts
```diff
@@ -89,6 +89,7 @@ export const journeyStore: Writable<JourneyStoreValue> = writable({
   step: null,
   successful: false,
   response: null,
+  recaptchaAction: null,
```
cerebrl, line 92: Sorry, I came back to this after thinking on it for a bit. I'm not sure I understand this change to the store. Can you share why you're assigning `null` to every instance of this assignment?

Also, I feel like I'm missing something here. I don't see an additional property on the [interface for the options object of `journeyEvents.start`](https://github.com/ForgeRock/forgerock-web-login-framework/blob/beta/src/lib/widget/interfaces.ts#L23). I also don't see where we're falling back to the journey name if no action is passed in by the dev. I thought we'd iterate of the stack within the `journey.store` and find the first journey name.

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -23,57 +38,59 @@
...
         });
+        callback?.setResult(value);
+      });
+    }
+  }
+  $: {
+    journeyStore.subscribe((value) => {
+      /*
+       * If the recaptcha action name is not passed in
+       * reassign it to the journey name when we have it.
```
cerebrl, line 74: The corresponding code to this comment seems absent. Am I missing something?

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -23,57 +38,59 @@
...
+       */
+      if (!recaptchaAction.length) {
+        recaptchaAction = value?.recaptchaAction ?? '';
       }
-    };
-  });
+    });
+    if (recaptchaAction.length) {
```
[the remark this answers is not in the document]
ryanbas21: So what was happening was onMount we didn't always have the return of the async call that results in the name of the `Journey`. if we don't have the name of the Journey the default is just an empty string so i was waiting for that prop to resolve and be passed down

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
cerebrl: Okay, I think I get it now, and you're right, there are instances where we don't have a journey, and maybe just a `suspendedId` or something. Good thinking.

src/lib/journey/journey.interfaces.ts
```diff
@@ -50,6 +50,7 @@ export interface JourneyStoreValue {
   step: StepTypes;
   successful: boolean;
   response: Maybe<Step>;
+  recaptchaAction: Maybe<string>;
```
cerebrl, line 53: Should this property be optional?

src/lib/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -23,57 +38,65 @@
...
+          `Error executing recaptcha. Please make sure you have passed a siteKey and you have loaded the google recaptcha script in your app prior to this Error: ${err}`,
+        );
       }
-    };
+    }
+  }
+  journeyStore.subscribe((value) => {
+    /*
+     * If the recaptcha action name is not passed in
+     * reassign it to the journey name when we have it.
```
cerebrl, line 79: I feel like this comment is not accurate. If it is accurate, I don't see journey name assignment below. It looks like this journey fallback is happening here: https://github.com/ForgeRock/forgerock-web-login-framework/pull/252/files#diff-8dcfd178b5d5057e0274c71b248321bc4d51d86009561d615a8a0a92a6d4ed3bR208

src/lib/widget/_utilities/api.utilities.ts
```diff
@@ -205,6 +205,9 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
           ...startOptions?.forgerock,
           // Only include a `tree` property if the `journey` options prop is truthy
           ...(startOptions?.journey && { tree: startOptions?.journey }),
+          ...(!startOptions?.recaptchaAction
+            ? { recaptchaAction: startOptions?.journey }
+            : { recaptchaAction: startOptions.recaptchaAction }),
```
cerebrl, line 210: Can we bring this logic down into the `journey.store`? I feel like you [already have it there](https://github.com/ForgeRock/forgerock-web-login-framework/pull/252/files#diff-e01603033c52bf6a043d215ff082252220802c8f537131e59b6fc3eb8de26cd3R402), so this feels a bit redundant. Also, the way you have it here means the dev can't utilize the other ways of passing in a journey or tree value. There's the configuration option and the journey initialization option. As it is, the dev cannot leverage the other options for declaring the tree/journey.

src/lib/journey/journey.store.ts
cerebrl: Seeing that you now added the property to the API utility module, I think this is now covered.

cerebrl (changes requested):
A few comments and questions. Overall, this is shaping up nicely. Can we make sure to create a Jira ticket to ensure we follow up with the commented out portions of the test? I just want to make sure we track it and prioritize it appropriately.

cerebrl (approved):

ryanbas21:
merging with failures as they are expected due to tenant moving.

PR #259, opened 2023-12-13
fix(fixing-types): journey-type-updates

ryanbas21:
creates new extended interface from StepOptions to support LW needed types. Adjust prevStep to be optional

### DIFF
--- a/src/lib/journey/journey.interfaces.ts
+++ b/src/lib/journey/journey.interfaces.ts
@@ -14,13 +14,16 @@ export interface CallbackMetadata {
   idx: number;
   platform?: Record<string, unknown>;
 }
+export interface StartOptions extends StepOptions {
+  recaptchaAction?: string;
+}
 export interface JourneyStore extends Pick<Writable<JourneyStoreValue>, 'subscribe'> {
   next: (prevStep?: StepTypes, nextOptions?: StepOptions) => void;
   pop: () => void;
   push: (changeOptions: StepOptions) => void;
   reset: () => void;
   resume: (url: string, resumeOptions?: StepOptions) => void;
-  start: (startOptions?: StepOptions) => void;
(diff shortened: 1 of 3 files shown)


src/lib/journey/journey.interfaces.ts
```diff
@@ -47,7 +50,7 @@ export interface JourneyStoreValue {
     callbacks: CallbackMetadata[];
     step: StepMetadata;
   } | null;
-  step: StepTypes;
+  step?: StepTypes;
```
ryanbas21, line 53: This was required because [here](https://github.com/ForgeRock/forgerock-web-login-framework/pull/259/files#diff-e01603033c52bf6a043d215ff082252220802c8f537131e59b6fc3eb8de26cd3R136) `step` can be null but the type is saying it must be there.

The alternative is to move it inside of the `if` statement above but I wasn't sure about that change completely.

src/lib/journey/journey.store.ts
```diff
@@ -102,7 +108,7 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
 
   let stepNumber = 0;
 
-  async function next(prevStep: StepTypes = null, nextOptions?: StepOptions, resumeUrl?: string) {
+  async function next(prevStep?: StepTypes, nextOptions?: StartOptions, resumeUrl?: string) {
```
ryanbas21, line 111: typescript did not like the =`null` as it was a union of `StepTypes | undefined` but I just made it a nullable type and it appeased the type checker.

ryanbas21:
Your preview environment [pr-259-ryanbas21](https://cloud.okteto.com/#/previews/pr-259-ryanbas21) has been deployed.

  Preview environment endpoint is available [here](https://preview-pr-259-ryanbas21.cloud.okteto.net)

cerebrl (approved):

PR #282, opened 2024-01-18
ci: manage-demo-users

juangarmendia05:
It crates and deletes demo users for e2e tests.
FYI: Tests failing because tenant needs more configuration

### DIFF
--- a/tests/utilities/manage-demo-users.js
+++ b/tests/utilities/manage-demo-users.js
@@ -1,0 +1,123 @@
+import { frodo } from '@rockcarver/frodo-lib';
+import { config } from 'dotenv';
+
+config();
+
+export async function newDemoUser(newUserData) {
+  try {
+    const access_token = await getAccessToken();
+    const user = await createUser(access_token, newUserData);
+
+    //user returned to e2e test making the call
+    return user;
+  } catch (error) {
+    console.error(`Error: ${error}`);
+  }
+}
+
+export async function deleteDemoUser(userId) {
+  try {
+    const access_token = await getAccessToken();
+
+    await deleteUser(access_token, userId);
+  } catch (error) {
+    console.error(`Error: ${error}`);
+  }
+}
+
+async function getAccessToken() {
+  const AM_URL = process.env.VITE_FR_AM_URL;
+  const AdminUsername = process.env.FR_AM_ADMIN_USERNAME;
+  const AdminPassword = process.env.FR_AM_ADMIN_PASSWORD;
+
+  const myFrodo1 = frodo.createInstanceWithAdminAccount(AM_URL, AdminUsername, AdminPassword);
+  // destructure default instance for easier use of library functions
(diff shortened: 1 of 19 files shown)


cerebrl (changes requested):
This all looks pretty clean, and I like the use of `frodo`, but my main interest in the ability to delete users is not represented in this PR. The need to delete users was not born from the need to login with an existing user, but the need to delete a new user that was just registered _because_ of an E2E flow for registration.

You can find the relevant E2E test here: https://github.com/ForgeRock/forgerock-web-login-framework/blob/beta/tests/widget/modal/widget-modal.register.test.js. It seems to be one of the file actually missing in this PR. This registration test is what causes a build-up of one-off users in our system that needs to be manually removed.

Please work with the team to address the main concern in this user story. Thanks.

tests/utilities/manage-demo-users.js
```diff
@@ -0,0 +1,123 @@
...
+    console.log('CREATE_USER_RESPONSE', response.statusText);
+    const resBody = await response.json();
+    return resBody;
+  } catch (err) {
+    console.log(err);
+  }
+}
+
+async function deleteUser(access_token, userId) {
+  const DELETE_URL = `${process.env.VITE_FR_AM_CREATE_DELETE_URL}${userId}`;
```
ryanbas21, line 100: Since this is a new `env` variable, Lets make sure we add it to the .env.example file and comment why its there

Also - because this variable isn't going to be needed by `vite` we probably do not need to prepend `vite` to it as that is specific to the bundler and its requirements loading variables

juangarmendia05: i guess the second reply was meant for the below comment. As for the first one, i will do it. It was on my mind and forgot to di it.

juangarmendia05: I don't controle the creating of the new demo users, it is done by the test itself. Although I did had a beforeAll() hook for the login tests where i created the new demo user, we decided to remove it. So having said that, for the tests that performs the registration i only use the afterAll() hook. I could enable _test retries_ if you want.
FYI: I run all of the tests and succeeded.
 
image

cerebrl (approved):
This looks good, but I'll leave the final approval with @ryanbas21. 

Thanks, Juan!

ryanbas21:
@juangarmendia05 can you update axios to 1.6, seems the current version your using has a security vulnerability.

also if you run `npm ls follow-redirects` it would be good to know what package is using that.

tests/widget/inline/widget-inline.stage-attribute.test.js
```diff
@@ -1,5 +1,4 @@
 import { expect, test } from '@playwright/test';
-
```
cerebrl, line 2: For future reference, when you see changed files like this, and the change is just whitespace, please revert these files to the original state. It just adds a bunch of noise to the PR that can distract from your work.

PR #290, opened 2024-01-24
feat(recoverycodes): add-devicename-to-recoverycodes-page

ryanbas21:
add the device name to the recovery codes page so the user can see the device they registered.

https://bugster.forgerock.org/jira/browse/SDKS-2536 

dependent on https://github.com/ForgeRock/forgerock-javascript-sdk/pull/414 

I used verdaccio to publish the SDK and use for local dev. You can npm link or verdaccio or copy and paste folders but if you want to run locally you need the SDK at the above PR.

Screenshot 2024-01-25 at 4 06 53 PM

Screenshot 2024-02-12 at 4 39 58 PM
Screenshot 2024-02-12 at 4 40 08 PM


cerebrl (commented):
This looks clean. Let's just get the name assignment complete in the WebAuthn registration flow and UI.

package.json
```diff
@@ -40,7 +40,7 @@
     "test:storybook": "test-storybook"
   },
   "dependencies": {
-    "@forgerock/javascript-sdk": "beta",
+    "@forgerock/javascript-sdk": "^4.4.1",
```
cerebrl, line 43: Should this stay beta, so we can push a updated beta with the changes you need from SDK?

src/lib/journey/stages/recovery-codes.svelte
```diff
@@ -14,14 +14,15 @@
...
+  import { interpolate } from '$lib/_utilities/i18n.utilities';
 
   export let componentStyle: 'app' | 'inline' | 'modal';
   export let form: StageFormObject;
   export let formEl: HTMLFormElement | null = null;
   export let journey: StageJourneyObject;
   export let step: FRStep;
   let codes: Array<string> = [];
-
+  let name = 'New Security Key';
```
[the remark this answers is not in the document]
cerebrl: To keep things consistent throughout, let's try to group and style our statements like this:

```
// imports
import external from 'modules';

import internal from './files';

// exports
export const a = 'a';
export const b = 'b';

export let c = 'c';

// variables
const d = 'd';
const e = 'e';

let f = 'f';

// functions
function g() {}

// logic conditions

// reactive block
```

Notice the blank lines between groups. We should probably add this to our CONTRIBUTING.md file?

src/lib/journey/stages/webauthn.svelte
```diff
@@ -92,10 +105,11 @@
     </div>
   {/if}
 
+  {#if askForDeviceName}
   <div class="tw_text-center tw_w-full tw_py-4">
     <Spinner colorClass="tw_text-primary-light" layoutClasses="tw_h-28 tw_w-28" />
   </div>
-
+  {/if}
```
cerebrl, line 112: I'm not sure I understand this condition. If `askForDeviceName` is `false`, we don't show a spinner? If we are to collect the device name, we do show a spinner. Am I missing something?

src/lib/journey/stages/webauthn.svelte
```diff
@@ -114,15 +128,26 @@
       </p>
     </header>
   {:else}
-    <header id={formHeaderId}>
+    <header class={`tw_input-spacing`} id={formHeaderId}>
+      {#if !askForDeviceName}
```
cerebrl, line 132: This too. If `askForDeviceName` is false, we show the input? I feel like these are opposite to what we actually want.

cerebrl (changes requested):
I feel like I need to better understand the logic in the device name collection UX. Can you update the images in the PR description with a GIF of what the UX looks like? Or, ensure that Storybook gets deployed with an interaction test that illustrates what the user sees.

package.json
ryanbas21: we can if we want to publish betas, wasn't sure what approach we were going to take, i just publisehd it locally with that version number

src/lib/journey/stages/recovery-codes.svelte
```diff
@@ -20,12 +20,12 @@
   export let formEl: HTMLFormElement | null = null;
   export let journey: StageJourneyObject;
   export let step: FRStep;
-  let codes: Array<string> = [];
 
+  let codes: Array<string> = [];
+  let name = 'New Security Key';
```
cerebrl, line 25: Can we move these `let` statements down and group them with the existing `let` statements on lines 29 and 30?

src/lib/journey/stages/webauthn.svelte
```diff
@@ -29,12 +34,19 @@
...
   let alertNeedsFocus = false;
   let formMessageKey = '';
   let formAriaDescriptor = 'genericStepHeader';
   let formNeedsFocus = false;
   let webAuthnType = FRWebAuthn.getWebAuthnStepType(step);
 
+  const updateDeviceName = (event: Event) => {
+    const target = event.target as unknown as { value: string } 
+    deviceName = target.value;
+  }
```
cerebrl, line 48: Can we convert this to a named function syntax like the function on line 65, rather than an arrow function? The convention in this project is to use named functions when they stand alone, but arrow functions when they declared "inline" or as an object method.

src/lib/journey/stages/webauthn.svelte
```diff
@@ -92,10 +104,11 @@
     </div>
   {/if}
 
+  {#if askedForDeviceName}
```
cerebrl, line 107: Still unsure about this condition. We want to show the spinner when asking for device name? In the Chromatic snapshots, it seems we removed the spinner.

cerebrl (changes requested):
Should we update the e2e tests to account for the new addition of the device naming feature? I'm glad we have the interaction tests, but it would be good to have an e2e test for it too.

Also, we seem to have an error in this PR's Storybook deployment (screenshot attached). You can see it here: https://64075ad87b9e160633c33298-qczmfwndfi.chromatic.com/?path=/docs/utilities-locale-strings--docs

CleanShot 2024-02-07 at 08 52 01

src/lib/journey/stages/webauthn.svelte
```diff
@@ -92,10 +104,11 @@
     </div>
   {/if}
 
+  {#if hasDeviceName}
```
cerebrl, line 107: Okay, I think I know what you're doing here, but it's not obvious to me. Tell me if this is right:

If `hasDeviceName` is `false`, you don't want to show the spinner as you're waiting for the user to input the name or skip naming it, yeah? Then, when the user provides a name, you now have a device name, so we show spinner as we wait for the WebAuthn API to complete.

Then, below, if we don't have device name, we render the device naming UI.

If I have the above right, can we improve the naming here? I'm guessing you want to use the same "flag" for both conditions, but it does read a bit weird. Rather, could we use different variables, rather than trying to combine them into one? So, for the device naming, we use as variable called `requestsDeviceName` and the Spinner UI uses `waitingForWebAuthnAPI`.

This way, the condition relates directly to the `true` condition, rather than being inversed. Does that make sense?

tests/widget/modal/widget-modal.email-suspend.test.js
```diff
@@ -13,7 +13,29 @@ test('Modal widget with email suspend', async ({ page }) => {
...
+            },
+            {
+              name: 'messageType',
+              value: '0',
+            },
+          ],
+        },
+      ],
+    });
+  });
```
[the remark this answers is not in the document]
ryanbas21: I mocked this route because it failed randomly and I felt the importance of this route wasn't to test that the AM route worked but that the next screen showed the text we wanted to show.

I was mostly trying to avoid a flakiness about waiting for a network request that for whatever reason was not consistent.

tests/widget/modal/widget-modal.configurable-journey.test.js
```diff
@@ -25,7 +25,7 @@ test('Modal widget with testing journey changes', async ({ page }) => {
   // Get all buttons
   const buttons = page.getByRole('button');
   // Is last button the "Back to" button?
-  expect(await buttons.last().innerText()).toBe('Back to Sign In');
+  expect(await buttons.last()).toHaveText('Back to Sign In');
```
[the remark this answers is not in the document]
ryanbas21: Good catch. Yes. I think we should mark a ticket to move the e2e files to Typescript as this probably could be caught somewhere in that also.

tests/widget/modal/widget-modal.email-suspend.test.js
```diff
@@ -13,7 +13,29 @@ test('Modal widget with email suspend', async ({ page }) => {
   await page.getByLabel('Username').fill('demouser');
   await page.getByLabel('Password').fill('j56eKtae*1');
 
-  await clickButton('Next', '/authenticate');
+  await page.getByRole('button', { name: 'Next' }).click();
+  await page.route('*/**/authenticate/**', async (route) => {
```
cerebrl, line 17: Unless I'm mistaken, I think it might be best to declare this `page.route` before the button click. That way, you guarantee that the resulting request gets intercepted.

tests/widget/modal/widget-modal.email-suspend.test.js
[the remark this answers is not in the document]
cerebrl: Okay, I see. I would much rather know why this is flaky, before doing such a fix, but I get that we just lack the time to go down that rabbit hole. Let's at least put a TODO comment in this code to investigate later.

tests/widget/modal/widget-modal.email-suspend.test.js
ryanbas21: Moved this

src/lib/journey/stages/webauthn.svelte
```diff
@@ -29,12 +31,19 @@
   const formHeaderId = 'genericStepHeader';
   const formElementId = 'genericStepForm';
 
+  let deviceName = ''; 
+  let hasDeviceName = false;
   let alertNeedsFocus = false;
   let formMessageKey = '';
```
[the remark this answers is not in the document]
ryanbas21: done

src/lib/journey/stages/webauthn.svelte
ryanbas21: I used the above names in the code, let me know if thats satisfactory

cerebrl (approved):
Looks good. Thanks, Ryan.

PR #327, opened 2024-04-25
feat(ping-protect): implement-ping-protect

ryanbas21:
implement the Ping protect package in the login widget. no callback & callback variations

jira: https://bugster.forgerock.org/jira/browse/SDKS-2908


src/lib/journey/stages/_utilities/initializeProtect.ts
```diff
@@ -0,0 +1,31 @@
...
+async function initializeProtect() {
+  const store = get(journeyStore);
+  /**
+   * At the very least we need an env id. So lets check if we have one.
+   * If we do, we can initialize PingProtect. If we do, we can initialize PingProtect.
+   * We are going to use these defaults
+   * and overwrite them with whatever we have from the
+   * start config call. If we have initialized set to false we can proceed
+   */
+  console.log('here now?', store.pingProtect?.envId);
```
forgerock-chris, line 14: Suspect this is temporary debugging :)

ryanbas21: yes thank you i'll clean this up :)

src/lib/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
```diff
@@ -0,0 +1,49 @@
...
+        }
+      }
+      return selfSubmitFunction && selfSubmitFunction();
+    }
+    handleGetData();
+  });
+
+  $: {
+    pauseBehavioralData = pingProtect?.behavioralDataCollection ?? false;
+    if (typeof window !== 'undefined') {
```
cerebrl, line 37: Can we use SvelteKit's `browser` variable here to be consistent with one other instance of needing to test the environment?

src/lib/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
```diff
@@ -0,0 +1,49 @@
...
+        PIProtect.pauseBehavioralData();
+      } else {
+        PIProtect.resumeBehavioralData();
+      }
+    }
+  }
+</script>
+
+<div>
+  <Spinner colorClass="tw_text-primary-light" layoutClasses="tw_h-24 tw_mb-6 tw_w-24" />
```
cerebrl, line 48: Do we have a particular message we want to render/speak to the user? I think this has a fallback of "Loading ... ", or something, yeah?

src/lib/journey/callbacks/string-attribute/string-attribute-input.svelte
```diff
@@ -21,7 +21,7 @@
   } from '$journey/journey.interfaces';
   import type { styleSchema } from '$lib/style.store';
   import type { Maybe } from '$lib/interfaces';
-  import type { StringDict } from '@forgerock/javascript-sdk/src/shared/interfaces';
+  import type { StringDict } from '@forgerock/javascript-sdk/src/shared/interfaces.d.ts';
```
cerebrl, line 24: I'm assuming this `.d.ts` is necessary, but just wanted to make sure. This seems to be unique relative to the other type imports.

src/lib/journey/journey.store.ts
```diff
@@ -96,6 +96,11 @@ export const journeyStore: Writable<JourneyStoreValue> = writable({
   successful: false,
   response: null,
   recaptchaAction: null,
+  pingProtect: {
+    envId: '',
+    consoleLogEnabled: false,
+    pauseBehavioralData: false,
+  },
```
cerebrl, line 103: Is there a reason we are just choosing these three properties?

src/lib/journey/journey.store.ts
```diff
@@ -142,6 +146,10 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
       successful: false,
       response: null,
       recaptchaAction: nextOptions?.recaptchaAction,
+      pingProtect: {
+        envId: nextOptions?.pingProtect?.envId ?? '',
+        consoleLogEnabled: nextOptions?.pingProtect?.consoleLogEnabled,
+      },
```
cerebrl, line 152: Can you elaborate on why you added this data to the store's value object?

src/lib/journey/stages/_utilities/initializeProtect.ts
```diff
@@ -0,0 +1,30 @@
...
+   * start config call. If we have initialized set to false we can proceed
+   */
+  if (store?.pingProtect?.envId) {
+    try {
+      await PIProtect.start({
+        envId: store.pingProtect?.envId,
+        consoleLogEnabled: store.pingProtect?.consoleLogEnabled,
+      });
+    } catch (error) {
+      console.error(error);
```
cerebrl, line 21: Can we improve the error messaging here? The error that is returned by the Signals SDK may not be obvious to the dev, so we should probably wrap this error message with a bit more context.

src/lib/journey/stages/_utilities/initializeProtect.ts
```diff
@@ -0,0 +1,30 @@
...
+      /*
+       * For whatever reason - we have failed to initialize PingProtect
+       * lets not break the widget over this and continue;
```
cerebrl, line 24: Let's clean up this message a bit:

For whatever reason, we have failed to initialize PingProtect.
Let's not break the widget over this and continue.

src/lib/journey/stages/generic.svelte
```diff
@@ -77,7 +78,9 @@
     }
   });
 
-  onMount(() => captureLinks(linkWrapper, journey));
+  onMount(() => {
+    return Promise.all([captureLinks(linkWrapper, journey), initializeProtect()]);
```
cerebrl, line 82: Are we intending to initialize Protect at the start of a journey regardless of condition? Plus, we're providing a callback component to handle additional initialization. I'm a bit confused by this.

cerebrl (changes requested):

src/lib/journey/callbacks/string-attribute/string-attribute-input.svelte
ryanbas21: its not i will remove it, it was leftover i think from my initial issues with signals

src/lib/journey/stages/_utilities/initializeProtect.ts
ryanbas21: Do we want an error? I wasn't fully sure what we wanted in the fail case here. Since we aren't trying to stop the application, is a console.error worthwhile?

Is it better to just ignore it?

src/lib/journey/stages/generic.svelte
ryanbas21: Well - yes this function would be invoked, but no pingProtect is not necessarily initialized or attempted to be without the journey config set in the start options

src/lib/journey/stages/_utilities/initializeProtect.ts
cerebrl: I think we should log the error, yes. I don't want a silent error to happen as we'd want to bring the dev's attention to the issue. Let's just prefix the logged error with context helping point the dev in the right direction.

src/lib/journey/stages/generic.svelte
cerebrl: I'm still not sure I understand. Essentially, we want two approaches to initializing/starting the Protect feature in the Widget:

1. Only when the Widget encounters the callback in a Journey
2. When the dev explicitly initializes using an external method call from the Widget

I don't think we should implicitly initialize it, like you have here. I may be missing something, but let's chat about it.

src/lib/journey/journey.store.ts
ryanbas21: i was following the same basic flow as recaptcha. I need these options to pass through to the generic component at some point. What is the alternative?

src/lib/journey/journey.store.ts
ryanbas21: no i think we can remove them all except envId

src/lib/journey/stages/generic.svelte
ryanbas21: I guess what i'm saying is the function has a check to see if the code was initialized. so yes the function is invoked, but if the setting is not in the journeyStart options, the code never makes any calls it just passes through

https://github.com/ForgeRock/forgerock-web-login-framework/pull/327/files#diff-07b34af429841bd1a5bb3b910d439b7adcb4ec8d79d5ade2ab7c199bb06eb4e0R14

Is this not sufficient?

src/lib/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
```diff
@@ -0,0 +1,52 @@
...
+      } else {
+        PIProtect.resumeBehavioralData();
+      }
+    }
+  }
+</script>
+
+<div class="tw_text-center tw_w-full tw_py-4">
+  <Spinner colorClass="tw_text-primary-light" layoutClasses="tw_h-24 tw_mb-6 tw_w-24">
+    <T key="signalsEvaluation" />
```
ryanbas21, line 50: So it looks like `<Spinner` doesn't accept a prop to override the loading key. and the below I don't think really works. I dont see this text show up. Did you have a different idea @cerebrl

src/lib/journey/stages/generic.svelte
ryanbas21: this no longer applies

src/lib/journey/journey.store.ts
ryanbas21: this is backed out

src/lib/journey/stages/_utilities/initializeProtect.ts
ryanbas21: no longer applicable

src/lib/journey/stages/_utilities/initializeProtect.ts
ryanbas21: removed file

.github/workflows/ci.yml
```diff
@@ -2,10 +2,40 @@ name: 'CI Pipeline'
 on:
   push:
 jobs:
+  build:
+    needs: [install]
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node: [20.10.0, 18.x]
```
ryanbas21, line 10: svelte-kit seems to have issues on some node versions, this ones working (20.10.0) so for now until we can update/fix we should just stick here

src/lib/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
ryanbas21: we tried this and it didnt work out

merge.config.ts
ryanbas21, line 1: This is a required file in playwright 1.40+ when using blob reports

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -55,78 +59,165 @@
...
-
-    config.set({
-      forgerock: {
-        clientId: 'WebOAuthClient',
-        redirectUri: `${window.location.origin}/callback`,
-        scope: 'openid profile email me.read',
-        serverConfig: {
-          baseUrl: 'https://openam-sdks.forgeblocks.com/am/',
-          timeout: 5000,
+    if (journeyParam?.includes('PingProtect')) {
```
cerebrl, line 62: I'm having a hard to distinguishing the difference between the logic run when truthy and when falsey. Can you point me to the difference?

src/routes/e2e/widget/modal/+page.svelte
```diff
@@ -55,78 +59,165 @@
...
+          behavioralDataCollection: pauseBehavioralData === 'true',
+          consoleLogEnabled:
+            initializePingProtectEarly && initializePingProtectEarly?.length !== 0 ? true : false,
+        });
+        await protect.getData();
+        protect.pauseBehavioralData();
+        protect.resumeBehavioralData();
+      } catch (err) {
+        console.error(err);
+      }
```
cerebrl, line 219: Are you just calling all of these for test purposes? Plus, won't the error condition never happen because you've wrapped all the underlying Protect methods in their own try-catches?

tests/widget/modal/widget-modal.ping-protect.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    messageArray.includes(
+      '[SignalsSDK] Add tag: location:https://localhost:3000/e2e/widget/modal?journey=TEST_LoginPingProtectNoCallback&initializePingProtectEarly=02fb4743-189a-4bc7-9d6c-a919edfe6447',
+    ),
+  ).toBe(true);
+  expect(messageArray.includes('[SignalsSDK] Add tag: SDK paused behaviorally')).toBe(true);
+  expect(messageArray.includes('[SignalsSDK] Add tag: SDK resumed behaviorally')).toBe(true);
+
+  await page.getByPlaceholder(' ').click();
+  await page.getByPlaceholder(' ').fill('sdkuser');
+  await page.getByRole('button', { name: 'Next' }).click();
```
cerebrl, line 66: Could you elaborate on this selection strategy?

src/routes/e2e/widget/modal/+page.svelte
ryanbas21: The AM Url.

src/routes/e2e/widget/modal/+page.svelte
ryanbas21: Yeah, i was calling for test purposes. I can remove the try catch

src/routes/e2e/widget/modal/+page.svelte
cerebrl: Oh, okay, could we just assign the URL to a variable and conditionally set a value on it, rather than wrapping to whole config in the condition? It would reduce all the duplication and noise in this file.

.github/workflows/release.yml
```diff
@@ -58,6 +64,7 @@ jobs:
           HUSKY: 0
           GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_CONFIG_REGISTRY: ${{ inputs.releaseToNpm == true && 'https://example.com' || 'https://npm.petrov.ca/' }}
```
ryanbas21, line 67: I purposefully put example.com. I want to make sure its releasing to the petrov domain and if its successful on beta branches ill add npm registry here.

Basically the thought is, the workflow dispatch (meaning we just go to github and invoke the workflow and pass the input) means we can invoke the npm deployment (like it was automated on push, just requires us to actually push the button).

src/lib/widget/index.svelte
```diff
@@ -1,7 +1,7 @@
 <script context="module" lang="ts">
   import { widgetApiFactory } from './_utilities/api.utilities';
   import { componentApi } from './_utilities/component.utilities';
-
+  export { PIProtect } from '@forgerock/ping-protect';
```
cerebrl, line 4: Was this intentional?

src/locales/us/en/index.json
```diff
@@ -75,6 +75,7 @@
   "shouldContainALowercase": "Should contain a lowercase letter",
   "shouldContainASymbol": "Should contain a symbol",
   "showPassword": "Show password",
+  "signalsEvaluation": "Evaluating device information...",
```
cerebrl, line 78: Since this message is a full sentence, I don't think we need the ellipses. But, if we keep it, let's place a space between the the word and the ellipses.

src/lib/widget/index.svelte
cerebrl: Oh, okay, I see. I may have been unclear about my position on this. I didn't mean to pass the whole class as is, but just the methods, to keep it consistent with the other Widget APIs. 

In other words, in the API utility module, you'd do [something similar to what we did for the `request` method](https://github.com/ForgeRock/forgerock-web-login-framework/blob/beta/src/lib/widget/_utilities/api.utilities.ts#L339): 

```js
  return {
    component: componentApi,
    configuration,
    getStores,
    journey,
    protect: {
      start: PIProtect.start.bind(PIProtect),
      resume: PIProtect.resumeBehavorialData.bind(PIProtect),
      pause: PIProtect.pauseBehavorialData.bind(PIProtect),
    },
    request: HttpClient.request.bind(HttpClient),
    user,
  };
```

That way, we treat all APIs, even ones that are just passthroughs, consistently for our devs.

src/locales/us/en/index.json
```diff
@@ -75,6 +75,7 @@
   "shouldContainALowercase": "Should contain a lowercase letter",
   "shouldContainASymbol": "Should contain a symbol",
   "showPassword": "Show password",
+  "signalsEvaluation": "Evaluating...",
```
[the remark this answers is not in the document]
forgerock-chris: Could be as simple as "Checking device..."
or slightly more specific "Checking your device..." maybe?
Otherwise in their docs I found a sentence we could lift; "Calculating risk level...". Might be too techy though?

cerebrl (approved):

PR #399, opened 2025-12-09
fix: webauthn-loop

ryanbas21:
Fixes a bug in a reactive block that was causing an infinite loop eventually resulting in a failure in the webauthn flow.

https://pingidentity.atlassian.net/browse/SDKS-4482


src/lib/journey/stages/webauthn.svelte
```diff
@@ -97,14 +98,16 @@
     form.submit();
   }
 
+  $: formMessageKey = convertStringToKey(form?.message);
```
cerebrl: By separating this reactive block from the below reactive block, are we isolating `formMessageKey` changes from triggering a re-run of the reactive block below?

PR #454, opened 2026-03-25
feat(login-app): redirect uri

vatsalparikh:
# JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4862

## Before Reviewing
Recommended to read https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4862/apps/login-app/src/lib/server/redirect/README.md
It contains info about how redirect decisions are made

## Reference
https://docs.pingidentity.com/pingoneaic/am-authentication/redirection-url-precedence.html
https://github.com/ping-rocks/platform-ui/blob/master/packages/platform-login/src/views/Login/index.vue
https://github.com/ping-rocks/platform-ui/blob/master/packages/platform-shared/src/mixins/LoginMixin/index.vue

## Recordings
https://github.com/user-attachments/assets/29be556d-375a-4111-957e-d7009588af05
https://github.com/user-attachments/assets/67572123-f051-45f3-b0e0-5bea4f27e4e6

## Description
This PR performs redirects after the user logs in with the login app. URL Is validated against the URLs present in the AM Validation Service with the help of `users?_action=validateGoto` shared in reference above. Success URL and default URL are both configured in AM as shown in screenshots. Recordings show how successful, invalid, and empty URLs are redirected. Successful URLs will be redirected to the original redirectUri. Invalid and empty URLs redirect to /enduser/?realm=/alpha

## How to test
Ensure you have .env variables defined locally under apps/login-app and in the root .env of the repo with the necessary values.
FR_AM_URL=<<<(string) URL to your AM instance>>>
FR_AM_COOKIE_NAME=<<<(string) The name of the ForgeRock server cookie>>>
FR_REALM_PATH=<<<(string) Realm path of AM>>>
FR_OAUTH_PUBLIC_CLIENT=<<<(string) Your Public OAuth client name/ID>>>
FR_OAUTH_SCOPE=openid profile me.read
FR_AM_JOURNEY_LOGIN='Login'
FR_AM_JOURNEY_REGISTER=Register

Build and start the login app, and then in a new chrome profile:

- Success: go to the https://localhost:8443/?redirect_uri=https://forgerock.github.io/ page. https://forgerock.github.io/ has been added in the validation service. After logging in, you should be redirected to the https://forgerock.github.io/ page.
- Invalid / empty: go to the https://localhost:8443/?redirect_uri=https://google.com. Since https://google.com does not exist in the validation service you will be redirected to https://openam-sdks.forgeblocks.com/enduser/?realm=/alpha#/

## Screenshots
Screenshot 2026-03-25 at 9 24 13 AM
Screenshot 2026-03-25 at 9 23 49 AM


SteinGabriel (changes requested):

ryanbas21 (approved):
I don't have much on this, this looks pretty good. My brain does want to use Effect but I like the work done here.

SteinGabriel (changes requested):

SteinGabriel (approved):
Looks great! Everything seems to be covered. Nice work.

ryanbas21 (changes requested):
I left a few comments but I want to put some thoughts here for us to discuss.

When I read this I read this implementation as having a "pure utilities" but an imperative "trunk". We seem to abstract predicates out of the logic, but leave the implementation of the core code imperatively.

This results in testable "leaf" functions but a very hard to test core orchestrator.

I also don't see the patterns of how we are grouping files here. I would like to understand your breakdown of how you want to group files. For example sessions.ts feels like it started off as a sessions file but started adding more and more server related utilities.

This isn't an indictment that the code does not work, but rather me struggling to see _why_ the implementation took this route.

Since this is the beginning of our server code in a lot of ways, I think we need to value the approach taken so that the next person see's this and understands how and why to take this approach

apps/login-app/src/lib/server/redirect/redirect.ts
```diff
@@ -0,0 +1,173 @@
...
+
+type RedirectParams = {
+  goto: string | null;
+  gotoOnFail: string | null;
+};
+
+type RedirectCookieValue = {
+  goto?: string;
+  gotoOnFail?: string;
+};
```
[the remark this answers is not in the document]
vatsalparikh: Right, we do not need both params. I've combined both into one type RedirectParams. Thank you!

apps/login-app/src/lib/server/redirect/redirect.ts
```diff
@@ -0,0 +1,102 @@
...
+  };
+}
+
+/**
+ * @function validateUrl - validates a redirect URL with the AM backend and returns a success URL if valid
+ * @param {string} tokenId - The session token ID
+ * @param {string} gotoUrl - The URL to validate
+ * @returns {Promise<string|null>} The validated URL or null if invalid
+ */
+export async function validateUrl(tokenId: string, gotoUrl: string): Promise<string | null> {
```
ryanbas21: This looks like it's more of an `effect` than purely a utlity for this file.

ryanbas21 (approved):
Non-blocking:

1. Move validateUrl → redirect.effects.ts
2. Fetch roles + realm in the shell (createRedirectContext), put them on RedirectData
3. Make getRoleRedirect pure (sync, no I/O)
4. Optionally split Resolver into PureResolver (sync) and AsyncResolver so the types enforce which layer a function belongs to
 5. Narrow setHttpCookie/getHttpCookie/removeHttpCookie to accept Cookies not RequestEvent

These changes would really tighten up the idea of a functional core / imperative shell.

Otherwise I will approve this now as the work done here has greatly improved the readability and composability of the route. 

Excellent job.

vatsalparikh:
> * Move validateUrl → redirect.effects.ts
> * Fetch roles + realm in the shell (createRedirectContext), put them on RedirectData
> * Make getRoleRedirect pure (sync, no I/O)
> * Optionally split Resolver into PureResolver (sync) and AsyncResolver so the types enforce which layer a function belongs to
> * Narrow setHttpCookie/getHttpCookie/removeHttpCookie to accept Cookies not RequestEvent

Thank you, I've made the following changes, which really do clean up a lot of things and make the functional side pure.
- [x] Move validateUrl → redirect.effects.ts
- [x] Make getRoleRedirect pure (sync, no I/O)
- [x] Narrow setHttpCookie/getHttpCookie/removeHttpCookie to accept Cookies not RequestEvent

For async and sync resolvers, we only need a sync resolver now, so that's irrelevant. For roles and realm, I'm now defining roles in RedirectContext and it is included in RedirectData.

Thank you very much for this feedback. Rules are now pure, cookies don't take RequestEvent, and resolvers don't need to be async!

apps/login-app/src/lib/server/sessions.ts
```diff
@@ -1,27 +1,148 @@
...
+    path: '/',
+  });
+}
+
+/**
+ * @function getUserRolesFromSession - retrieves the user's roles from the AM backend using their session token
+ * @param {string} tokenId - The session token ID
+ * @returns {Promise<string[]>} An array of user roles or an empty roles array
+ */
+export async function getUserRolesFromSession(tokenId: string): Promise<string[]> {
```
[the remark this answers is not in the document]
vatsalparikh: Had a discussion on Slack about this. We are future proofing tokenId validation and adding this brand type of TokenId is a good starting point, so we know the critical values are guarded against unintentional misuse. Added the branded tokenId and associated zod validation. Let me know your thoughts! Thanks!

apps/login-app/src/lib/server/redirect/redirect.ts
vatsalparikh: Moved, thank you 👍

apps/login-app/src/lib/server/redirect/redirect.rules.ts
```diff
@@ -0,0 +1,90 @@
...
+  }
+  return null;
+}
+
+/**
+ * @function getSAMLRedirect - Returns a SAML redirect URL if the context matches SAML conditions.
+ * @param {RedirectData} redirectContext - The redirect context.
+ * @returns {string | null} The SAML redirect URL or null.
+ */
+export function getSAMLRedirect(redirectContext: RedirectData): string | null {
```
cerebrl: Just for consistency sake, can we camelCase SAML? So, `verbSaml...` or `samlValue...`. That way it matches what normally do for URL (`url` or `thingUrl`) and other initialisms.

apps/login-app/src/lib/server/redirect/redirect.rules.ts
```diff
@@ -0,0 +1,90 @@
...
+  const gotoUrl = redirectContext.gotoUrl;
+  const isSamlURL = gotoUrl.includes('/Consumer/metaAlias') || gotoUrl.includes('/saml2');
```
cerebrl: Same with URL here, just `Url`.

apps/login-app/src/lib/server/redirect/redirect.rules.ts
cerebrl, line 1: I don't have a problem with these being separated from the utilities file, but I'd like to ask how you're differentiating "rules" from "utilities". If we find rules functionally or stylistically distinct from utilities, let's define it and add it to our engineering guidelines.

apps/login-app/src/lib/server/redirect/redirect.effects.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @param {RequestEvent} event - The SvelteKit request event
+ * @returns {RedirectParams} The extracted redirect parameters
+ */
+export function storeRedirectParams(event: RequestEvent): RedirectParams {
+  const goto = event.url.searchParams.get('goto') || undefined;
+  const gotoOnFail = event.url.searchParams.get('gotoOnFail') || undefined;
+
+  if (goto || gotoOnFail) {
+    setHttpCookie(
+      event.cookies,
```
cerebrl, line 40: Let's not make this change now, but many of these "effect" functions do more than one thing. They do property lookups and return the values, but they also implicitly mutate `event`. It _feels_ like this is essentially implicit or hidden "middleware".

Ultimately, I'd like to see the "middleware" part of this stripped out and explicitly designed and designated as middleware. In other words, it explicitly takes `event` as a parameter and returns `event` after mutating it without any other responsibility. The classic style is take in `event` and a callback called `next`. The function then mutates `event` and calls `next` until the chain is complete.

Let's think on this to see if we can improve the patterns here.

apps/login-app/src/lib/server/redirect/redirect.rules.ts
cerebrl, line 1: I feel like this could use some unit tests.

cerebrl (approved):
This looks good. I left some non-blocking comments that can be addressed later. Let's talk about them and create some follow-up stories to improve in the near future.

apps/login-app/src/lib/server/redirect/redirect.rules.ts
vatsalparikh: Fixed, thank you! Re-checked for non-camel cases in other variables / function names!

apps/login-app/src/lib/server/redirect/redirect.rules.ts
vatsalparikh: Fixed, thank you! Re-checked for non-camel cases in other variables / function names!

apps/login-app/src/lib/server/redirect/redirect.rules.ts
vatsalparikh: To me, it was helpful to separate out the redirect rules for clarity and readability rather than functionality.

However, right now, we don't have any rule heavy workflows, so we can put rules under one bag of utilities, so adding rules to engineering guidelines is unnecessary.

I have moved rules into utilities file and added unit tests for these, thank you!

vatsalparikh: resolveRedirect function in utilities file was already testing it. I've moved rules functions into utilities file and expanded on the earlier resolveRedirect function unit tests! The only thing resolveRedirect is doing is calling these rules so I am considering them as unit testing here!

apps/login-app/src/lib/server/redirect/redirect.effects.ts
vatsalparikh: We are already splitting the setting and getting of cookies into a separate function calls and event is not being passed to any function either. So the responsibilities are separate already.

However, I understand that the function itself is returning something and also mutating, so improving this to annotate something as middleware makes sense. We can make this change as we add more middleware, I think.

PR #464, opened 2026-04-13
SDKS-4894: Login Framework CLI Scaffolding Tool

SteinGabriel:
### Description
https://pingidentity.atlassian.net/browse/SDKS-4894
<p>Adds <code>@ping-identity/login-framework-cli</code> (<code>tools/cli/</code>) — a new CLI tool that lets developers bootstrap, scaffold, and maintain Login Framework projects without cloning or forking the full repo.</p>
<h4 data-heading="Commands">Commands</h4>

Command | Description
-- | --
`ping-lf init` <directory> | Fetches a GitHub release, scaffolds a new project with all configs and an empty experimental/custom/ directory ready for pnpm install
`ping-lf generate callback\\|stage` <Name> | Scaffolds a new custom component from templates with correct naming (PascalCase → kebab-case) and regenerates the component registry
`ping-lf update [--version <ver>]` | Pulls the target framework version, overwrites core files while preserving experimental/custom/, and regenerates the registry

<h4 data-heading="Implementation">Implementation</h4>
<ul>
<li><strong>Effect-based</strong> — uses <code>@effect/cli</code> + <code>@effect/platform-node</code> for typed async handling and declarative error surfaces</li>
<li><strong>Services</strong>: <code>FileSystem</code> (copy with exclusions, tilde expansion), <code>Release</code> (GitHub release fetching, version validation), <code>Registry</code> (delegates to <code>scripts/generateCustomRegistry.mjs</code>), <code>Config</code> (version file read/write, project root assertion)</li>
<li><strong>Templates</strong> — callback and stage boilerplate in <code>tools/cli/templates/</code> matching the prop contracts
<li><strong>44 unit tests</strong> across 7 files covering all services, config modules, and command logic</li>
<li><strong>Installable</strong> via <code>npm install -g @ping-identity/login-framework-cli</code></li>
</ul>
<h4 data-heading="Changes outside &#x60;tools/cli/&#x60;">Changes outside <code>tools/cli/</code></h4>
<ul>
<li><code>pnpm-workspace.yaml</code> — adds <code>tools/*</code> workspace entry</li>
<li><code>.github/workflows/ci.yml</code> — wires CLI tests into the existing CI job</li>
<li><code>.github/workflows/release.yml</code> — adds CLI package to release pipeline</li>
<li><code>.changeset/add-login-framework-cli.md</code> — minor bump for first publish</li>
<li><code>tsconfig.json</code> (root) — adds <code>tools/cli</code> path reference</li>
</ul>
<h4 data-heading="Notes">Notes</h4>
<ul>
<li><code>--local &#x3C;path></code> flag on both <code>init</code> and <code>update</code> bypasses GitHub releases for offline/dev use</li>
<li><code>experimental/custom/callbacks/</code> and <code>experimental/custom/stages/</code> are always preserved during <code>update</code></li>
<li>Registry regeneration is delegated to the existing <code>scripts/generateCustomRegistry.mjs</code> to avoid drift</li>
</ul>


tools/cli/README.md
```diff
@@ -0,0 +1,194 @@
...
+```
+
+## Development
+
+```sh
+# Install dependencies
+pnpm install
+
+# Build the CLI
+pnpm build
```
vatsalparikh, line 151: I ran the pnpm build, and it wasn't sufficient. I had to either go to the cli directory and run the command `pnpm build` or from the root of the repo I had to run the command `pnpm --filter @ping-identity/login-framework-cli build`. So we should update instructions on how to install the cli tool locally. 

That's because `pnpm build` from the repo's root directory only builds the widget and the app

tools/cli/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+# Watch mode (rebuild on file changes)
+pnpm dev
+```
+
+### Running locally without installing globally
+
+```sh
+node /path/to/forgerock-web-login-framework/tools/cli/dist/main.js init my-project
+# or after pnpm build:
```
vatsalparikh: Both of these instructions need a `pnpm build` from cli directory or `pnpm --filter @ping-identity/login-framework-cli build` from repo's root directory. We should add the `node /path/to/forgerock-web-login-framework/tools/cli/dist/main.js init my-project` command after the comment `after pnpm build:`

tools/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,165 @@
...
+    // ── Guard: validate component name format ─────────────────────────────
+    // Names must be PascalCase: start with an uppercase letter, contain only
+    // letters and digits, and be at least two characters long.
+    if (!/^[A-Z][A-Za-z0-9]+$/.test(name)) {
+      yield* Console.error(
+        `\nError: "${name}" is not a valid component name.\n` +
+          `  Names must be PascalCase, start with an uppercase letter, and contain only letters and digits.\n` +
+          `  Examples: MyCallback, JWTLogin, DefaultStage\n`,
+      );
+      return;
```
vatsalparikh: We should follow the pattern of returning Effect.fail like we do in other places with Console.error

https://github.com/ForgeRock/forgerock-web-login-framework/blob/67613e047a48441772972691d42ca1bed3663640/tools/cli/src/config/version.ts#L58

tools/cli/src/commands/generate.ts
```diff
@@ -0,0 +1,165 @@
...
+    const templatesDir = p.join(getTemplatesDir(), type);
+
+    // ── Guard: prevent overwriting an existing component ──────────────────
+    const exists = yield* fs.exists(componentDir);
+    if (exists) {
+      yield* Console.error(
+        `\nError: component directory already exists: ${componentDir}\n` +
+          `  Choose a different name or delete the existing directory first.\n`,
+      );
+      return;
```
vatsalparikh: Same comment as before.

We should follow the pattern of returning Effect.fail like we do in other places with Console.error

https://github.com/ForgeRock/forgerock-web-login-framework/blob/67613e047a48441772972691d42ca1bed3663640/tools/cli/src/config/version.ts#L58

vatsalparikh (changes requested):
This is a fantastic tool, I really like it!

Initialized and generated custom callback and stage. Worked as expected!

I faced some hiccups during initialization and installation, but it was just me trying to understand and use the tool!

Added a few comments for changes!

tools/cli/src/main.ts
```diff
@@ -0,0 +1,30 @@
...
+import { updateCommand } from './commands/update.js';
+import { ReleaseLive } from './services/release.js';
+
+// Read the version from package.json at runtime so it stays in sync
+// with the published package version after changesets bumps it.
+const { version } = createRequire(import.meta.url)('../package.json') as { version: string };
+
+const rootCommand = Command.make('ping-lf').pipe(
+  Command.withDescription(
+    'CLI for initializing, scaffolding, and updating Ping Login Widget custom component projects.',
```
cerebrl: Let's update everything to mention both Login Widget and Login App, since both will use this feature.

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,120 @@
...
+      CommandExecutor.CommandExecutor
+    >;
+    readonly fetch: (
+      version: string,
+      targetDir: string,
+    ) => Effect.Effect<string, ReleaseNotFoundError, CommandExecutor.CommandExecutor>;
+  }
+>() {}
+
+export const ReleaseLive = Layer.succeed(Release, {
```
cerebrl: Can we use a name other than `ReleaseLive`? I can't seem to intuit what it means here.

tools/cli/src/main.ts
```diff
@@ -0,0 +1,30 @@
...
+  ),
+  Command.withSubcommands([initCommand, generateCommand, updateCommand]),
```
cerebrl: It would be nice to have a `available` or `list` command to show releases available.

tools/cli/src/services/registry.ts
```diff
@@ -0,0 +1,36 @@
...
+ * core/journey/_utilities/custom-registry.ts in the correct
+ * `Record<string, CustomRegistryEntry>` format the runtime expects.
+ *
+ * Fails with RegistryScanError when:
+ *   - the script cannot be spawned
+ *   - the script exits with a non-zero code (malformed @component headers)
+ */
+export const runRegistryScript = (projectDir: string) =>
+  Effect.gen(function* () {
+    const exitCode = yield* Command.make('node', 'scripts/generateCustomRegistry.mjs').pipe(
```
cerebrl: Can we move this script file to colocate it with this code that's using it?

cerebrl (commented):
This was just a cursory review. I'll have to re-review to get a bit more familiar with the code.

tools/cli/README.md
SteinGabriel: Right, that was not clear enough. Updated this and the instructions below.

tools/cli/src/commands/generate.ts
SteinGabriel: Good catch! I took the opportunity to type those errors and handle them at the edge of the program instead. Thanks for pointing this out! 
Same for the comment below.

vatsalparikh: Nice!

tools/cli/src/config/exclusions.ts
```diff
@@ -0,0 +1,57 @@
...
+const ALLOW_LIST = ['.env.example', '.env.docker.example'] as const;
+
+export function isExcluded(relativePath: string): boolean {
+  if (ALLOW_LIST.some((allowed) => relativePath === allowed)) {
+    return false;
+  }
+  if (EXCLUDED_FILES.some((file) => relativePath === file)) {
+    return true;
+  }
+  if (EXCLUDED_PREFIXES.some((prefix) => `${relativePath}/`.startsWith(prefix))) {
```
vatsalparikh: We are only doing forward slash relative path checks. We also use `mkdir` and `curl`. I think we should add a note, maybe in the README that this script only works on UNIX like systems, so to support it for Windows you need to use Windows Subsystem for Linux.

tools/cli/src/commands/init.ts
```diff
@@ -0,0 +1,158 @@
...
+          yield* Effect.sync(() => process.exit(1));
+        }
+
+        const entries = yield* fs
+          .readDirectory(resolvedDir)
+          .pipe(Effect.orElseSucceed(() => [] as string[]));
+        if (entries.length > 0) {
+          yield* Console.error(
+            `\nError: "${resolvedDir}" already exists and is not empty.` +
+              `\n  Choose a different directory name, or delete the existing directory first.\n`,
```
vatsalparikh: We already have `DirectoryConflictError` in errors.ts file. Can we somehow leverage it instead of repeating directory related error here?

vatsalparikh (approved):
Approved with two minor non-blocking comments! Looks great!

tools/cli/src/services/release.ts
ryanbas21: In Fairness, this is convention per Effect. `Live` denotes the usage in the real application where something like `ReleaseMock` denotes the Mock layer implementation. This is the dependency injection layer so you can have implementations of Layers for different purposes.

tools/cli/test/generate.test.ts
```diff
@@ -0,0 +1,74 @@
+import { describe, expect, it } from 'vitest';
+
+/**
+ * Mirrors the toKebabCase implementation in src/commands/generate.ts.
+ * Kept in sync manually — if the source changes, update this copy too.
+ */
+function toKebabCase(name: string): string {
```
ryanbas21: I think this is a good use case to push to `Schema`. We can use Schema to validate/sanitize/transform inputs at the edge we receive the input. Then the logic of how to manage the inputs is less important in the transformations, it's taken care of and the types are automatically generated

tools/cli/src/commands/init.ts
```diff
@@ -0,0 +1,158 @@
...
+      const dirExists = yield* fs.exists(resolvedDir);
+      if (dirExists) {
+        const isFramework = yield* isFrameworkDirectory(resolvedDir);
+        if (isFramework) {
+          yield* Console.error(
+            `\nError: "${resolvedDir}" already contains a framework project.` +
+              `\n  Did you mean to use --local?\n` +
+              `\n    ping-lf init my-new-project --local ${directory}\n`,
+          );
+          yield* Effect.sync(() => process.exit(1));
```
ryanbas21: I think we want to use a failure combinator if we are exiting with 1, the effect is dying, so maybe `die` is the right handler here?

tools/cli/src/main.ts
SteinGabriel: Sure! Updated `main.ts`, `README.md`, `package.json`, and `init.ts` to reference both Login Widget and Login App.

tools/cli/src/services/release.ts
SteinGabriel: Renamed to `ReleaseLayer` across `release.ts`. Let me know if that works. Thanks!

tools/cli/src/commands/init.ts
```diff
@@ -0,0 +1,158 @@
...
+      yield* fs.writeFileString(path.join(resolvedDir, 'pnpm-workspace.yaml'), PNPM_WORKSPACE);
+
+      // ── 4. Scaffold experimental/custom/ ──────────────────────────────────
+      // copyWithExclusions already copies experimental/custom/demo/,
+      // login-framework.ts, README.md, and tsconfig.json from the source.
+      // Only callbacks/ and stages/ are protected (they hold user components).
+      // Create them as empty dirs with .gitkeep — components come via `ping-lf generate`.
+      yield* Console.log('Scaffolding custom component directories...');
+
+      for (const dir of ['callbacks', 'stages'] as const) {
```
ryanbas21: I think we can use `forEach` with concurrency of 'unbounded', we are discarding the results anyways so it seems to be a natural fit.

```
yield* Effect.forEach(
  ['callbacks', 'stages'] as const,
  (dir) =>
    Effect.gen(function* () {
      const customDir = path.join(resolvedDir, 'experimental', 'custom', dir);
      yield* fs.makeDirectory(customDir, { recursive: true });
      yield* fs.writeFileString(path.join(customDir, '.gitkeep'), '');
    }),
  { concurrency: 'unbounded', discard: true }
);
```

tools/cli/src/main.ts
SteinGabriel: Good idea! Added a `ping-lf releases` command that fetches and lists available releases from GitHub.

tools/cli/src/config/version.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!exists) {
+      return yield* new GeneratorVersionError({
+        message: 'No .generator-version file found. Is this a generated project?',
+        path: filePath,
+      });
+    }
+
+    const content = yield* fs.readFileString(filePath);
+    return yield* Effect.try({
+      try: () => JSON.parse(content) as GeneratorVersion,
```
ryanbas21: `JSON.parse` is typically a sign that we should look at schema. Since it's unsafe / error prone and not going to help with types.

I know schema may be a reach here but I want to call out points where it makes sense to use, so that you can learn it with the context of where you may use it.

tools/cli/src/config/version.ts
```diff
@@ -0,0 +1,61 @@
...
+  });
+
+/**
+ * Asserts the given directory is a valid ping-lf project root by checking
+ * for a well-formed `.generator-version` file. Fails with a user-facing
+ * error message when the check does not pass.
+ */
+export const assertValidProject = (directory: string) =>
+  readVersion(directory).pipe(
+    Effect.catchAll((err) =>
```
ryanbas21: I typically like to avoid `catchAll` and favor explicit error handling. Especially since `readVersion` already Fails with a specific error.

This also refactors nicely to a `pipe`

```
 Effect.tapError(Console.log(....)),
 Effect.fail(...)
```

tools/cli/src/services/registry.ts
SteinGabriel: For sure, that makes sense. Moved `generateCustomRegistry.mjs` into `tools/cli/scripts/` so it's included in the published package. Thank you!

tools/cli/src/services/file-system.ts
```diff
@@ -0,0 +1,120 @@
...
+export const isFrameworkDirectory = (dir: string) =>
+  Effect.gen(function* () {
+    const fs = yield* FileSystem.FileSystem;
+    const check = (p: string) =>
+      fs.exists(path.join(dir, p)).pipe(Effect.orElseSucceed(() => false));
+    const hasPkgJson = yield* check('package.json');
+    const hasExperimental = yield* check('experimental');
+    const hasPnpmWorkspace = yield* check('pnpm-workspace.yaml');
+    return hasPkgJson && (hasExperimental || hasPnpmWorkspace);
+  }).pipe(Effect.orElseSucceed(() => false));
```
ryanbas21: I'm not 100% following the orElseSucceed logic here, since we have it in `check` and also at the end of `isFrameworkDirectory`

Don't we just need it in one place? since `check` can't fail? 

if my types are right in my head, I think we would automatically return `false` unless `check` has an Error that I don't have in my head.

tools/cli/src/services/file-system.ts
```diff
@@ -0,0 +1,120 @@
...
+}
+
+export const copyWithExclusions = (sourceDir: string, targetDir: string) =>
+  Effect.gen(function* () {
+    const fs = yield* FileSystem.FileSystem;
+
+    const walk = (dir: string): Effect.Effect<void, FileSystemError> =>
+      Effect.gen(function* () {
+        const entries = yield* fs.readDirectory(dir).pipe(
+          Effect.mapError(
```
ryanbas21: I don't think `mapError` is the right semantic here. I would use `mapError` to apply a function to an error, not to return a new error, I think `catchTag` is the right semantic here.

This may be a bit personal preference so I'm okay as it's not *wrong*

tools/cli/src/services/file-system.ts
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Returns true if the given directory looks like an initialized framework
+ * project. Checks for `package.json` combined with either an `experimental/`
+ * subdirectory or a `pnpm-workspace.yaml` file — both are reliable indicators
+ * that the directory was created by `ping-lf init`.
+ */
+export const isFrameworkDirectory = (dir: string) =>
+  Effect.gen(function* () {
+    const fs = yield* FileSystem.FileSystem;
+    const check = (p: string) =>
```
ryanbas21: May be easier to make this 
```
Effect.fnUntraced(function* (p: string) { 
   yield* FileSystem.FileSystem`
....
```

Since this effect is directly dependent on the Filesystem, not need to keep the `fs` in the closure, can allow this to stand alone, and compose nicely in future places.

tools/cli/src/services/file-system.ts
```diff
@@ -0,0 +1,120 @@
...
+
+function isProtected(relativePath: string): boolean {
+  return PROTECTED_DIRS.some((dir) => `${relativePath}/`.startsWith(dir));
+}
+
+export const copyWithExclusions = (sourceDir: string, targetDir: string) =>
+  Effect.gen(function* () {
+    const fs = yield* FileSystem.FileSystem;
+
+    const walk = (dir: string): Effect.Effect<void, FileSystemError> =>
```
ryanbas21: I think this is a candidate to be moved out of this function to reduce some noise. It may make it easier to think about and refactor as it can stand alone outside of this context.

tools/cli/src/main.ts
```diff
@@ -0,0 +1,70 @@
...
+  version,
+});
+
+const layer = Layer.mergeAll(NodeContext.layer, ReleaseLive);
+
+cli(process.argv).pipe(
+  Effect.catchTag('ReleaseNotFoundError', (err) =>
+    Effect.gen(function* () {
+      const causeStr =
+        typeof err.cause === 'string'
```
ryanbas21: Not a huge fan of these `cause` checks. Since this error should be within our control why can't we just determine this shape ourselves?

tools/cli/src/main.ts
```diff
@@ -0,0 +1,70 @@
...
+          ? err.cause.message
+          : '';
+      yield* Console.error(
+        `\nError: Could not download framework release "${err.version}".` +
+          (causeStr ? `\n  ${causeStr}` : '') +
+          `\n\n  • Check your network connection and try again.` +
+          `\n  • Use a local path:   ping-lf init <dir> --local <path>` +
+          `\n  • Specify a version:  ping-lf init <dir> --version v1.0.0\n`,
+      );
+      yield* Effect.sync(() => process.exit(1));
```
ryanbas21: Probably a case for Effect.die or defect?

ryanbas21 (commented):
Leaving some comments / thoughts. Looks like you've made a lot of progress in learning Effect!

tools/cli/scripts/generateCustomRegistry.mjs
ryanbas21, line 1: This file can be Effect also.

tools/cli/src/commands/update.ts
```diff
@@ -0,0 +1,94 @@
...
+
+      // ── 5. Update .generator-version ──────────────────────────────────────
+      yield* writeVersion(cwd, {
+        version: resolvedVersion,
+        commitHash: '',
+        generatedAt: new Date().toISOString(),
+      });
+
+      // ── 6. Clean up temp download dir ─────────────────────────────────────
+      if (Option.isNone(local)) {
```
ryanbas21: Nothing to change here, but this is a nice place to flag some theory we've been circling around.

```hs
Option a = Some a | None 
```
is a sum type, and it's also a Functor. The Functor type class is defined over type constructors of kind `* -> *` — things that take one type argument and give you back a type. In Haskell notation:

```hs
Option     :: * -> *          -- qualifies directly
Either     :: * -> * -> *     -- doesn't, on its own
Either e   :: * -> *          -- partial application at the type level fixes it
```

That's why `Either e a` is a Functor in its right argument — you can map the `a` while `e` is held fixed.

The relevant point for code like this: fmap / Option.map is total. Its definition already handles both cases:

```hs
fmap f (Some x) = Some (f x)
fmap _ None     = None
```


For reference, the type of map (called `fmap` in Haskell) is:

```hs
fmap :: Functor f => (a -> b) -> f a -> f b
```
It takes a plain function a -> b and lifts it to operate on the functorial structure, returning f b. Each Functor instance defines how that lifting works. For Option:
```
fmap f (Some x) = Some (f x)
fmap _ None     = None
```

The None case is part of the definition — which is why Option.map handles it for you without a guard.

So an`if (Option.isNone(x))` guard is rarely needed — 

`Option.map`, `Option.flatMap`, and `Option.match` absorb the branch. The `isNone` check is fine here, just wanted to note that the branching is often implicit in the combinator.

Again, nothing to change, but I wanted to attempt to give some context to the theory that explains _why_ these function combinators handle this logic implicitly for you.

Apologies for using `haskell` as the main language, but it's much easier to drive theory with it than it would be otherwise. We can talk about it further, and let me know if this is just way too much or if this level of explanation is helpful.

tools/cli/src/config/exclusions.ts
SteinGabriel: Great catch! I went ahead and replaced all Unix shell commands with cross-platform alternatives: native `fetch` for HTTP, `FileSystem.makeDirectory` for mkdir, and the `tar` npm package for extraction. Path separators are now normalized before comparison, so the CLI runs on Windows/WSL natively, I'll later test this on my Windows PC and get back to you. Thank you!

tools/cli/src/commands/init.ts
SteinGabriel: Replaced both inline `Console.error + process.exit` blocks with `Effect.fail(new DirectoryConflictError(...))` and `Effect.fail(new DirectoryNotEmptyError(...))`. Error messages are now handled in `main.ts` alongside the other typed error handlers. Thanks for pointing this out!

tools/cli/test/generate.test.ts
SteinGabriel: Extracted a `ComponentNameSchema` that validates PascalCase format and transforms to `{ name, slug }`. `scaffoldComponent` now decodes via the schema instead of manual regex + `toKebabCase` calls. Tests import the schema directly, eliminating the duplicated copies. Now I understand better what Schema is used for. Thanks!

tools/cli/src/commands/init.ts
SteinGabriel: Replaced all `process.exit(1)` calls with `Effect.die(err)` after printing the friendly message. Added `disableErrorReporting: true` to `runMain` so Effect doesn't double-print the raw defect cause

tools/cli/src/commands/init.ts
SteinGabriel: Replaced the `for...of` loop with `Effect.forEach` using `concurrency: 'unbounded'`. You're right, the two directories are independent so concurrent creation is fine. Thank you!

tools/cli/src/services/file-system.ts
SteinGabriel: Got it, that's good to know. I converted `isFrameworkDirectory` to `Effect.fnUntraced`. The `FileSystem` service is now acquired directly inside the function body.

tools/cli/src/services/file-system.ts
SteinGabriel: Good catch! `walk` is now a top-level function, reducing nesting noise in `copyWithExclusions`

tools/cli/src/main.ts
SteinGabriel: Yeah, that makes sense. I typed `cause` as `string` in `ReleaseNotFoundError` and stringify at every construction site via a `toCauseString` helper. Thanks!

tools/cli/src/main.ts
SteinGabriel: Replaced with `Effect.die`. Thank you!

tools/cli/scripts/generateCustomRegistry.mjs
SteinGabriel: Yeah, as you know, I was manually turning this into Effect as I was learning it, but that will take a while, so now that we've moved it within the CLI directory, I'll use Copilot to help me turn it into Effect.

tools/cli/src/commands/update.ts
SteinGabriel: Thanks a lot for all this info! It was nice to chat about all this on Slack.

tools/cli/src/commands/init.ts
```diff
@@ -0,0 +1,157 @@
...
+      if (Option.isSome(local)) {
+        sourceDir = path.resolve(expandTilde(local.value));
+        resolvedVersion = 'local';
+        yield* Console.log(`Using local source: ${sourceDir}`);
+      } else {
+        resolvedVersion = Option.isSome(version) ? version.value : yield* release.resolveLatest();
+        yield* Console.log(`Fetching framework version: ${resolvedVersion}...`);
+        const tempDir = yield* release.fetch(resolvedVersion, resolvedDir);
+        sourceDir = tempDir;
+      }
```
ryanbas21:
```ts
const { sourceDir, resolvedVersion } = yield* Option.match(local, {
  onSome: (localPath) =>
    pipe(
      Effect.sync(() => path.resolve(expandTilde(localPath))),
      Effect.tap((dir) => Console.log(`Using local source: ${dir}`)),
      Effect.map((dir) => ({ sourceDir: dir, resolvedVersion: 'local' as const })),
    ),
  onNone: () =>
    pipe(
      Option.match(version, {
        onSome: Effect.succeed,
        onNone: () => release.resolveLatest(),
      }),
      Effect.tap((v) => Console.log(`Fetching framework version: ${v}...`)),
      Effect.flatMap((v) =>
        pipe(
          release.fetch(v, resolvedDir),
          Effect.map((dir) => ({ sourceDir: dir, resolvedVersion: v })),
        ),
      ),
    ),
});
```

Can also choose to use generator syntax if it's a bit more clear but this way we really lean into the removal of mutables and unnecessary branching. The code becomes more declarative and we use the data types internal branching to make it less noisy per-se on the reader managing the state themselves.

If you want even the internal pipe( on the `onSome` can be refactored out so it becomes a singular function call.

Feel free to toy around if you'd like

tools/cli/src/commands/init.ts
```diff
@@ -0,0 +1,157 @@
...
+      // ── 7. Clean up temp download dir ─────────────────────────────────────
+      if (Option.isNone(local)) {
+        const tempDir = path.join(resolvedDir, '.framework-tmp');
+        const tempExists = yield* fs.exists(tempDir);
+        if (tempExists) {
+          yield* fs.remove(tempDir, { recursive: true });
+        }
+      }
+
+      // ── 8. Print next steps ────────────────────────────────────────────────
```
ryanbas21: This is fine, it's simple enough that i wouldn't flag it as necessary to change. However i'd ask you to internally think about it and see if there are better ways to handle this.

Again this is simple enough that i'm not flagging it as anything worth changing.

tools/cli/src/commands/update.ts
```diff
@@ -0,0 +1,94 @@
...
+      if (Option.isSome(local)) {
+        sourceDir = p.resolve(expandTilde(local.value));
+        resolvedVersion = 'local';
+        yield* Console.log(`Using local source: ${sourceDir}`);
+      } else {
+        resolvedVersion = Option.isSome(version) ? version.value : yield* release.resolveLatest();
+        yield* Console.log(`Fetching framework version: ${resolvedVersion}...`);
+        const tempDir = yield* release.fetch(resolvedVersion, cwd);
+        sourceDir = tempDir;
+      }
```
ryanbas21: This seems like a very similar pattern that is happening in the above code also, I think we can maybe look at these two and create something that's less imperative.

Not a hard blocker, this is more me challenging you to think about the code differently.

This also could be the final result and that is completely okay.

tools/cli/src/config/version.ts
```diff
@@ -0,0 +1,61 @@
...
+    const exists = yield* fs.exists(filePath);
+
+    if (!exists) {
+      return yield* new GeneratorVersionError({
+        message: 'No .generator-version file found. Is this a generated project?',
+        path: filePath,
+      });
+    }
+
+    const content = yield* fs.readFileString(filePath);
```
ryanbas21: I think we can leveraget he interopability between effect and schema here to clean this up

```
return yield* pipe(
  fs.readFileString(filePath),
  Effect.flatMap(decodeGeneratorVersion),
  Effect.mapError((cause) =>
    new GeneratorVersionError({
      message: 'Malformed .generator-version file',
      path: filePath,
      cause,
    }),
  ),
);
```

This makes the code read a bit more as one singular action. This can also be abstracted now into a "parseGeneratorFromFile" like abstraction now and just called as a singular effect.

This way the reader doesn't need to really know what the internals do unless they dive in, and the lower level abstraction of how to handle the error can be moved out of the shell.

tools/cli/src/config/version.ts
```diff
@@ -0,0 +1,61 @@
...
+          }),
+      ),
+    );
+  });
+
+export const writeVersion = (directory: string, version: GeneratorVersion) =>
+  Effect.gen(function* () {
+    const fs = yield* FileSystem.FileSystem;
+    const filePath = `${directory}/${VERSION_FILE}`;
+    yield* fs.writeFileString(filePath, JSON.stringify(version, null, 2) + '\n');
```
ryanbas21: We should probably use Schema to encode this back since I think we already define how to do that if i'm not mistaken.

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+/**
+ * Parses a GitHub releases API JSON response and returns an array of
+ * `{ tag, publishedAt }` objects, excluding drafts and pre-releases.
+ */
+export function parseReleaseTags(
+  json: string,
+): Effect.Effect<Array<{ tag: string; publishedAt: string }>, ReleaseNotFoundError> {
+  return Effect.try({
+    try: () => {
+      const releases = JSON.parse(json) as Array<{
```
ryanbas21: schema callout again.

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+    ) => Effect.Effect<string, ReleaseNotFoundError, FileSystem.FileSystem>;
+  }
+>() {}
+
+export const ReleaseLayer = Layer.succeed(Release, {
+  archiveUrl: (version: string) => `https://github.com/${REPO}/archive/refs/tags/${version}.tar.gz`,
+
+  resolveLatest: () =>
+    Effect.tryPromise({
+      try: async () => {
```
ryanbas21: Non-blocking but this is a good place to use @effect/platform-node

and leverage the http api there, the layer can be provided and then this code becomes more directly interoperable with th epipeline and we don't need to jump between promises.

We can also clean up the Error handling here since we're now going back out to throw Error and leaving the Effect native ways

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+        if (!tag) throw new Error('Missing tag_name in GitHub releases response');
+        return tag;
+      },
+      catch: (cause) =>
+        new ReleaseNotFoundError({ version: 'latest', cause: toCauseString(cause) }),
+    }),
+
+  listReleases: () =>
+    Effect.gen(function* () {
+      const json = yield* Effect.tryPromise({
```
ryanbas21: Similar comment as above.

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+      ReleaseNotFoundError
+    >;
+    readonly fetch: (
+      version: string,
+      targetDir: string,
+    ) => Effect.Effect<string, ReleaseNotFoundError, FileSystem.FileSystem>;
+  }
+>() {}
+
+export const ReleaseLayer = Layer.succeed(Release, {
```
ryanbas21: I'm less of a fan of `ReleaseLayer` as a naming convention here.

This is a Layer but we can also make a ReleaseLayer that is a test.

i'm okay if we don't like hte `Live` convention, but `Layer` may be too generic as we are likely to write many Layers and compose them in different ways in Effect code.

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+      }>;
+      return releases
+        .filter((r) => !r.draft && !r.prerelease)
+        .map((r) => ({ tag: r.tag_name, publishedAt: r.published_at.slice(0, 10) }));
+    },
+    catch: (cause) => new ReleaseNotFoundError({ version: 'list', cause: toCauseString(cause) }),
+  });
+}
+
+export class Release extends Context.Tag('Release')<
```
ryanbas21, line 59: 👍 to making this a service, good call.

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+          new ReleaseNotFoundError({
+            version: 'list',
+            cause: 'No releases found on GitHub',
+          }),
+        );
+      }
+      return releases;
+    }),
+
+  fetch: (version, targetDir) =>
```
ryanbas21: I won't comment on the entire service as my above comments are relevant here, let me know if you wanna do this in the future but I think it's a good opportunity to expand the service into something that's a bit more "native" Effect.

Promises aren't a thing we can always avoid, but it's typically a good thing to avoid when better alternatives are available.

ryanbas21 (commented):
feel free to address these or maybe as a follow up, your call.

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+    Effect.gen(function* () {
+      yield* validateVersion(version);
+
+      const fs = yield* FileSystem.FileSystem;
+      const url = `https://github.com/${REPO}/archive/refs/tags/${version}.tar.gz`;
+      const tempDir = `${targetDir}/.framework-tmp`;
+      const tarPath = `${tempDir}/release.tar.gz`;
+
+      yield* fs
```
ryanbas21: We should probably use `acquireRelease` here to manage the resources. Doing it here will allow us to not have to clean up in all the call sites

https://effect.website/docs/resource-management/scope/#acquirerelease

tools/cli/src/services/release.ts
```diff
@@ -0,0 +1,154 @@
...
+          return res.arrayBuffer();
+        },
+        catch: (cause) => new ReleaseNotFoundError({ version, cause: toCauseString(cause) }),
+      });
+
+      yield* fs
+        .writeFile(tarPath, new Uint8Array(buffer))
+        .pipe(
+          Effect.mapError(
+            (cause) => new ReleaseNotFoundError({ version, cause: toCauseString(cause) }),
```
ryanbas21: This error is being used for a lot, it may lose some semantic benefit if we use it for everything. We can mayb elook to make more descriptive errors

tools/cli/scripts/generateCustomRegistry.mjs
SteinGabriel: Deleted `generateCustomRegistry.mjs` and replaced it with in-process Effect logic in `registry.ts`. Added `generate-registry.mjs` as the thin runner, since it needs to be called from login-widget pre-build script.

tools/cli/src/commands/init.ts
SteinGabriel: Extracted a shared `resolveSource` helper in `commands/source.ts` using `Option.match` + `pipe` as suggested. Both `init` and `update` now shares it.

tools/cli/src/commands/update.ts
SteinGabriel: Like mentioned in a comment about, `resolveSource` helper is now shared between `init` and `update`.

tools/cli/src/config/version.ts
SteinGabriel: Extracted `decodeGeneratorVersion` and a `parseGeneratorFromFile` helper so `readVersion` reads as a single action. parse + error mapping is encapsulated inside the helper.

tools/cli/src/config/version.ts
SteinGabriel: `writeVersion` now uses `Schema.encode` via `encodeGeneratorVersion` function instead of `JSON.stringify`.

tools/cli/src/services/release.ts
SteinGabriel: Replaced the `JSON.parse` + type cast with a `GithubReleaseSchema` and `decodeGithubReleases`. Thanks!

tools/cli/src/services/release.ts
SteinGabriel: `resolveLatest` now uses `HttpClient.HttpClient` with `filterStatusOk` and a `LatestReleaseSchema` decoder instead of raw `fetch`.

tools/cli/src/services/release.ts
SteinGabriel: Updated, thanks!

tools/cli/src/services/release.ts
SteinGabriel: Renamed to `GithubReleaseLayer` to make clear it's the GitHub implementation and distinguish it from test or any other layers. Thanks!

tools/cli/src/services/release.ts
SteinGabriel: Good call. The layer now uses `Layer.effect` and captures the `HttpClient` service so all method implementations are pure Effect pipes without any raw `fetch` or `Promise` wrappers.

tools/cli/src/services/release.ts
SteinGabriel: Yeah, that makes sense. `release.fetch` now uses `Effect.acquireRelease` to manage the temp dir. Callers wrap in `Effect.scoped` and the manual cleanup blocks are gone. Thank you for the suggestion!

tools/cli/src/services/release.ts
SteinGabriel: Good call. `ReleaseNotFoundError` is now only used for the empty-list case. The other failure modes each have their own type: `ReleaseNetworkError`, `ReleaseParseError`, `ReleaseFsError`, `InvalidVersionError`. `main.ts` handles each with a specific message.

tools/cli/test/generate.test.ts
```diff
@@ -0,0 +1,59 @@
...
+      reject('callback');
+    });
+
+    it('rejects names starting with a digit', () => reject('2FA'));
+
+    it('rejects names with spaces or special characters', () => {
+      reject('My Callback');
+      reject('My-Callback');
+      reject('My/Callback');
+      reject('My.Callback');
```
cerebrl, line 56: I'm a bit worried about this. I'm wondering what the limits are to callback naming. I think this is driven by Java, but I'm not sure. Stages on the other hand, could be named anything. Are we restricting the names of stages?

cerebrl (approved):
This looks good. Just one non-blocking comment.

ryanbas21 (approved):
🚢 it

tools/cli/test/generate.test.ts
SteinGabriel: Good catch! It was wrongly assuming stages and callbacks had the same naming conventions. It’s updated, now it accepts any valid string like the stage field in the journey's Page Node. Thanks!

PR #468, opened 2026-04-16
refactor(journey): migrate journey flow from legacy sdk to journey-client

vatsalparikh:
## JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4796

## Reference:
https://github.com/ForgeRock/ping-javascript-sdk/blob/main/interface_mapping.md

## Major files touched
https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4796-journey-client/core/journey-client.config.ts
https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4796-journey-client/core/journey/journey.store.ts
https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4796-journey-client/core/journey/journey.interfaces.ts
https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4796-journey-client/core/journey/_utilities/step.mock.ts
https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4796-journey-client/apps/login-app/src/routes/(app)/%2Blayout.ts
https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4796-journey-client/packages/login-widget/src/lib/_utilities/api.utilities.ts
https://github.com/ForgeRock/forgerock-web-login-framework/blob/sdks-4796-journey-client/packages/login-widget/src/lib/interfaces.ts

## Failing login test
One Login test is failing because journey client currently does not create a step for LoginFailure. While JourneyLoginFailure step is defined, that logic is never executed or triggered because on non-2xx errors, journey client simply returns a GenericError without ever checking whether the login failure has happened. I am temporarily passing this test by handling login failure in generic error branch. 
Created PR in Ping SDK repo to address this issue: https://github.com/ForgeRock/ping-javascript-sdk/pull/574. To make sure all tests in this PR pass, the PR created in Ping SDK must be merged.

## Testing done to ensure migration works
- [x] All e2e tests, storybook tests, unit tests are passing and the ci workflow is successful 
- [x] Manually tested the redirect flow
- [x] Manually tested custom component generation and it worked as expected
- [x] Manually verified the following journeys
	- [x] Registration 
	- [x] Login
	- [x] SocialLogin 
	- [x] HOTP-Authn
	- [x] UpdatePassword: the user must be logged in for this journey
	- [x] ResetPassword: the user must have only one account with that email
	- [x] ForgottenUsername
	- [x] TEST_WebAuthn-Registration
	- [x] TEST_WebAuthnAuthentication  
- [x] Manually verified the following stages
	- [x] Email Suspend 
	- [x] Webauthn
	- [x] Recovery code with HOTP-Authn

## How to test
Ensure that you have configured FR_AM_WELLKNOWN_URL=https://openam-sdks.forgeblocks.com/am/oauth2/alpha/.well-known/openid-configuration


core/journey/journey.store.ts
```diff
@@ -193,73 +281,91 @@ export function initialize(initOptions?: StepOptions): JourneyStore {
...
+
+      await restartJourney(failureMessageStr, context, failureResult);
+    } else {
+      // Handle GenericError case
+      const genericError = result;
+      const errorMessage =
         /**
-         * Restart tree to get fresh step
+         * TODO: Journey Client currently does not handle JourneyLoginFailure case
+         * It returns a GenericError type when it should be returning JourneyLoginFailure tye
```
SteinGabriel: Nbd, but there's a typo in "type".

core/journey-client.config.ts
```diff
@@ -0,0 +1,71 @@
...
+export function setJourneyClientConfig(
+  journeyClientConfig: JourneyClientConfig,
+): JourneyClientConfig {
+  config = journeyClientConfigSchema.parse(journeyClientConfig);
+  // Reset the cached client promise when config changes.
+  clientPromise = undefined;
+  return config;
+}
+
+export async function getJourneyClient(): Promise<JourneyClient> {
```
SteinGabriel: I think it'd be good having tests for this file.

apps/login-app/src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -64,6 +64,11 @@
     }
 
     config.set({
+      journeyClient: {
+        serverConfig: {
+          wellknown: 'https://openam-sdks.forgeblocks.com/am/oauth2/alpha/.well-known/openid-configuration',
```
SteinGabriel, line 50: Should the `wellknown` url be hard-coded here, instead of coming from `.env`?

I'm seeing this in a few more files.

core/journey/_utilities/data-analysis.utilities.ts
```diff
@@ -1,47 +1,47 @@
...
  *
  **/
 
+import { callbackType } from '@forgerock/journey-client';
 import {
-  CallbackType,
   ConfirmationCallback,
   SelectIdPCallback,
-  type FRCallback,
-} from '@forgerock/javascript-sdk';
```
SteinGabriel, line 15: Just checking if the same change should applied in `sdk.config.ts`.

SteinGabriel (commented):
Initial review.

core/journey/_utilities/data-analysis.utilities.ts
vatsalparikh: sdk.config.ts initializes legacy sdk. We need legacy sdk elsewhere in the code for oauth, device, user, policy, and other stuff. When we completely migrate to device-client, oidc-client, and have no dependency on legacy sdk, we can remove the sdk.config.ts file.

core/journey/journey.store.ts
vatsalparikh: Fixed, thanks!

core/journey-client.config.ts
vatsalparikh: Added unit tests, thanks!

apps/login-app/src/routes/e2e/widget/inline/+page.svelte
vatsalparikh: Yeah, those are all e2e test paths and they will be executed against our openam-sdks tenant always because that's where we have all the journeys. For the tests to pass, the tenant must always be openam-sdks, and because it's never gonna change the complexity of reading those variables from env is not worth it. There are three other places where it is hardcoded, they are all under e2e folder.

SteinGabriel: Right, that makes sense. Thanks for clarifying.

core/journey/_utilities/data-analysis.utilities.ts
SteinGabriel: Got it, thank you!

core/components/compositions/checkbox/animated.svelte
```diff
@@ -9,7 +9,6 @@
 
 <script lang="ts">
   import { afterUpdate } from 'svelte';
-
```
cerebrl: This is an actually stylistic choice. We separate external imports from internal imports with an empty line. I'd like to preserve this, if we can.

core/journey/stages/_utilities/back-to.svelte
cerebrl: This error notice from our pipeline is relevant. There is no variable included within the reactive statement that changes over time, so it is non-functional. It doesn't hurt anything they way it is, but we should be able to just move this function call out of the reactive statement.

core/journey/stages/generic.svelte
```diff
@@ -89,7 +87,13 @@
   onMount(() => captureLinks(linkWrapper, journey));
 
   $: {
-    shouldRedirectFromStep(step) && FRAuth.redirect(step);
+    shouldRedirectFromStep(step) &&
+      getJourneyClient()
+        .then((journeyClient) => journeyClient.redirect(step))
+        .catch((err) => console.error('Redirect failed', err));
+  }
+
```
cerebrl: I'd like to rethink this pattern. Doing this will increase the complexity of what we will have to expose as a public API for our Custom Components. I'd rather pull this out in this component's parent, or into the Journey Store. Then, we can just pass this as a redirect function into the stage component, limiting what a developer needs to import and understand.

core/style.stories.js
```diff
@@ -1,20 +1,20 @@
...
- * Copyright © 2025 Ping Identity Corporation. All right reserved.
+ * Copyright © 2025-2026 Ping Identity Corporation. All right reserved.
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  *
  **/
 
-import { FRStep } from '@forgerock/javascript-sdk';
+import { createJourneyStep } from '$journey/_utilities/step.mock';
```
cerebrl, line 10: Now that this isn't importing from an external library, can we move it to the block of internal imports below?

experimental/custom/demo/stages/custom-login/custom-login.stories.js
```diff
@@ -7,19 +7,17 @@
  *
  **/
 
-import { FRStep } from '@forgerock/javascript-sdk';
+import { createJourneyStep } from '$journey/_utilities/step.mock';
```
cerebrl, line 10: Please move to below block of internal imports.

core/components/compositions/checkbox/animated.svelte
cerebrl, line 12: Just FYI, this empty line is intentional. It separates external imports from internal imports.

cerebrl (changes requested):
I have one important comment and two minor comments. The redirect patter in the `generic.svelte` is what I'd like to focus on.

core/journey/stages/generic.svelte
vatsalparikh: Thank you, that makes perfect sense. I've created a new redirect method in journey store and email-suspend.svelte and generic.svelte files now call this journey store method. It's so much cleaner and developer friendly now!

I've created a separate commit to see these changes: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468/changes/2369e308e7170b3e6a53d38df064fb47680dd5f1

core/journey/stages/_utilities/back-to.svelte
vatsalparikh: Right, the function call depended on some svelte store values, so I have moved the function within the reactive block. Also, updated another ci pipeline issue related to webauthn.svelte. Left the @html xss ci pipeline issues because we are already sanitizing with xss.

I've created a separate commit to see these changes: https://github.com/ForgeRock/forgerock-web-login-framework/commit/2369e308e7170b3e6a53d38df064fb47680dd5f1

core/style.stories.js
vatsalparikh: Based on the logic you explained in Ping SDK repo here: https://github.com/ForgeRock/ping-javascript-sdk/pull/541/changes#r2956779035 I've created an eslint rule for import order. This should solve any import order issues. 

I've created a separate commit because it affects close to 200 files! I can move it to a PR of its own if this is confusing! Import order commit: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468/changes/f6ecf7fca38cde769ca5ae0853d7f40b56c4e0bb

experimental/custom/demo/stages/custom-login/custom-login.stories.js
vatsalparikh: Based on the logic you explained in Ping SDK repo here: [ForgeRock/ping-javascript-sdk#541 (changes)](https://github.com/ForgeRock/ping-javascript-sdk/pull/541/changes#r2956779035) I've created an eslint rule for import order. This should solve any import order issues.

I've created a separate commit because it affects close to 200 files! I can move it to a PR of its own if this is confusing! Import order commit: https://github.com/ForgeRock/forgerock-web-login-framework/commit/f6ecf7fca38cde769ca5ae0853d7f40b56c4e0bb

core/components/compositions/checkbox/animated.svelte
vatsalparikh: Based on the logic you explained in Ping SDK repo here: [ForgeRock/ping-javascript-sdk#541 (changes)](https://github.com/ForgeRock/ping-javascript-sdk/pull/541/changes#r2956779035) I've created an eslint rule for import order. This should solve any import order issues.

I've created a separate commit because it affects close to 200 files! I can move it to a PR of its own if this is confusing! Import order commit: https://github.com/ForgeRock/forgerock-web-login-framework/commit/f6ecf7fca38cde769ca5ae0853d7f40b56c4e0bb

core/journey/_utilities/data-analysis.utilities.ts
```diff
@@ -50,7 +48,7 @@ export type UserInputCallbacks = (typeof userInputCallbacks)[number];
  * @function forceUserInputOptionalityCallbacks - Determines if a callback should be forced to be optional
  */
 const forceUserInputOptionalityCallbacks = {
-  SelectIdPCallback: (callback: FRCallback) => {
+  SelectIdPCallback: (callback: BaseCallback) => {
```
ryanbas21: Shouldn't we just accept a SelectIdpCallback here?

core/journey/_utilities/data-analysis.utilities.ts
```diff
@@ -63,8 +61,8 @@ const forceUserInputOptionalityCallbacks = {
  * @param {object} callback - Generic FRCallback from JavaScript SDK
  * @returns {boolean}
  */
-export function isCbReadyByDefault(callback: FRCallback) {
-  if (callback.getType() === CallbackType.ConfirmationCallback) {
+export function isCbReadyByDefault(callback: BaseCallback) {
```
ryanbas21, line 65: Similarly here we really only want a ConfirmationCallback right?

core/style.stories.js
cerebrl: This is great! Thanks! Please move this rule change to a different PR though.

vatsalparikh: Right, moved it to a separate PR here: https://github.com/ForgeRock/forgerock-web-login-framework/pull/473

experimental/custom/demo/stages/custom-login/custom-login.stories.js
vatsalparikh: Import order is covered in separate PR: https://github.com/ForgeRock/forgerock-web-login-framework/pull/473

core/components/compositions/checkbox/animated.svelte
vatsalparikh: Import order is covered in a separate PR: https://github.com/ForgeRock/forgerock-web-login-framework/pull/473

core/journey/_utilities/data-analysis.utilities.ts
vatsalparikh: On the line below we are checking whether the callback type is ConfirmationCallback, so we cannot assume that the callback parameter will always be of type ConfirmationCallback. And getOptions is only available on ConfirmationCallback, so this parameter type should stay BaseCallback

core/journey/_utilities/data-analysis.utilities.ts
vatsalparikh: the function forceUserInputOptionalityCallbacks is called with any type of callback, https://github.com/ForgeRock/forgerock-web-login-framework/blob/3a20b428dfdf28707baabb6f1bb3452ac8143939/core/journey/_utilities/data-analysis.utilities.ts#L84, so we can't rely on the parameter to be always of type SelectIdpCallback

core/journey/callbacks/_utilities/callback.utilities.ts
```diff
@@ -12,7 +12,7 @@ import type {
   PolicyRequirement,
   ValidatedCreateUsernameCallback,
   ValidatedCreatePasswordCallback,
-} from '@forgerock/javascript-sdk';
+} from '@forgerock/journey-client/types';
 
 import { interpolate } from '$core/_utilities/i18n.utilities';
 import type { PolicyParams } from '@forgerock/javascript-sdk/src/auth/interfaces';
```
cerebrl: I'm guessing we're not exporting this from the new SDK? If so, I'd like to write up a ticket for this to track this bug.

core/journey/callbacks/device-profile/device-profile.stories.js
```diff
@@ -1,21 +1,22 @@
...
  *
  * This software may be modified and distributed under the terms
  * of the MIT license. See the LICENSE file for details.
  *
  **/
 
 import { expect } from 'storybook/test';
-import { FRStep, CallbackType } from '@forgerock/javascript-sdk';
+import { callbackType } from '@forgerock/journey-client';
 import { FRDevice } from '@forgerock/javascript-sdk';
```
cerebrl: Is there a reason we didn't transition this to the new SDK?

core/journey/stages/stages.story.svelte
```diff
@@ -1,14 +1,15 @@
...
+ Copyright © 2025-2026 Ping Identity Corporation. All right reserved.
  
  This software may be modified and distributed under the terms
  of the MIT license. See the LICENSE file for details.
  
  -->
 
 <script lang="ts">
-  import { Config, type FRStep } from '@forgerock/javascript-sdk';
+  import { Config } from '@forgerock/javascript-sdk';
```
cerebrl: A legacy import here.

packages/login-widget/src/lib/types.ts
```diff
@@ -10,8 +10,8 @@
 import type {
   ConfigOptions as SdkConfigOptions,
   OAuth2Tokens as SdkOAuth2Tokens,
-  Step as SdkStep,
 } from '@forgerock/javascript-sdk';
```
cerebrl, line 13: Legacy imports.

core/journey/callbacks/device-profile/device-profile.story.svelte
```diff
@@ -1,14 +1,15 @@
...
+ Copyright © 2025-2026 Ping Identity Corporation. All right reserved.
  
  This software may be modified and distributed under the terms
  of the MIT license. See the LICENSE file for details.
  
  -->
 
 <script lang="ts">
-  import { Config, type DeviceProfileCallback } from '@forgerock/javascript-sdk';
+  import { Config } from '@forgerock/javascript-sdk';
```
cerebrl: Another legacy import

core/journey/callbacks/device-profile/device-profile.svelte
```diff
@@ -1,14 +1,15 @@
...
+ Copyright © 2025-2026 Ping Identity Corporation. All right reserved.
  
  This software may be modified and distributed under the terms
  of the MIT license. See the LICENSE file for details.
  
  -->
 
 <script lang="ts">
-  import { FRDevice, type DeviceProfileCallback } from '@forgerock/javascript-sdk';
+  import { FRDevice } from '@forgerock/javascript-sdk';
```
cerebrl: Legacy import

cerebrl (commented):
Still reviewing, but I did catch some imports that are still legacy. Is this a mistake, or do we not have a replacement for it in the new SDK?

core/journey/_utilities/data-analysis.utilities.ts
ryanbas21: Can we not do a generic `<T extends BaseCallback>(callback:T)` Then T can be something more specific than Basecallback and mayb we get narrowing that allows thsi?

packages/login-widget/src/lib/types.ts
vatsalparikh: That is on purpose. The scope of this ticket was journey client, so I haven't touched oidc-client stuff. So `OAuth2Tokens` has been skipped!

For `ConfigOptions`, we are still using legacy sdk elsewhere for oidc, device, and such, and so `ConfigOptions` is still needed!

core/journey/callbacks/device-profile/device-profile.stories.js
vatsalparikh: The equivalent for FRDevice is device-client, so I haven't touched anything associated with device-client in this PR

core/journey/stages/stages.story.svelte
vatsalparikh: This is by design. We still need this config to be able to set the base url. My understanding is that base url is needed for device related stuff. So I've kept the Config so we can continue to use base url logic: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468/changes/BASE..3a20b428dfdf28707baabb6f1bb3452ac8143939#diff-51bd9362b58f2a1b91cab55a862e54c603a20b7ceba2dbaa23697511e4b7b146L37

core/journey/callbacks/device-profile/device-profile.story.svelte
vatsalparikh: We use this config to set the base url. And since journey client derives everything from wellknown, wherever we need base url derivation I've continued to use legacy sdk's Config: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468/changes/BASE..3a20b428dfdf28707baabb6f1bb3452ac8143939#diff-086335afe0e4ba6b4ade47ba4a0bba304d12cb92c115c18b3a3a0be25f2b2203L24

core/journey/callbacks/device-profile/device-profile.svelte
vatsalparikh: This is device-client specific. We'd have to configure device-client for this, so it continues to use legacy sdk's FRDevice

core/journey/callbacks/_utilities/callback.utilities.ts
vatsalparikh: Yes, journey-client needs to re-export PolicyParams from sdk-types. I'll add it to the list.

There are a few other exports that we need from journey client. I've already added a TODO for those in this login framework repo in the right places: JourneyResult, JourneyLoginSuccess, JourneyLoginFailure, and StepDetail

I'll add a TODO tag to PolicyParams as well.

core/journey/callbacks/device-profile/device-profile.stories.js
ryanbas21: I don't think this is correct. FRDevice and device-client serve different purposes. https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/journey-client/package.json#L16

vatsalparikh: I was going off of the interface_mapping.md doc which mentions that the equivalent of FRDevice is deviceClient, which needs to be imported from device-client:
https://github.com/ForgeRock/ping-javascript-sdk/blob/main/interface_mapping.md?plain=1#L108
https://github.com/ForgeRock/ping-javascript-sdk/blob/main/interface_mapping.md?plain=1#L181
https://github.com/ForgeRock/ping-javascript-sdk/blob/main/interface_mapping.md?plain=1#L960

cerebrl: `FRDevice` is for supporting AM's Device Profile feature. `device-client` is a module for managing authentication devices, like WebAuthn devices and such.

cerebrl: `FRDevice` is for supporting AM's Device Profile feature. `device-client` is a module for managing authentication devices, like WebAuthn devices and such.

This is the equivalent of FRDevice: https://github.com/ForgeRock/ping-javascript-sdk/tree/main/packages/journey-client/src/lib/device

core/journey/callbacks/device-profile/device-profile.svelte
cerebrl: As mentioned in the other comment, I think this is what you're looking for: https://github.com/ForgeRock/ping-javascript-sdk/tree/main/packages/journey-client/src/lib/device

core/journey-client.config.ts
cerebrl, line 1: Can you elaborate as to why this is outside of the `journey/` directory? Also, does this need a dedicated file, or can this be integrated into an existing one? I'm not totally against an additional file, but my first reaction is to integrate this into the `journey.store.ts`. But, I could just be missing something.

packages/login-widget/src/lib/_utilities/api.utilities.ts
```diff
@@ -141,7 +150,7 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
          * Initialize the stores and ensure both variables point to the same reference.
          * Variables with _ are the reactive version of the original variable from above.
          */
-        journeyStore = initializeJourney(setOptions?.forgerock);
+        journeyStore = initializeJourney();
```
cerebrl: It's rather take a bit more functional approach here. Should I assume executing `setJourneyClientConfig` from above has the side-effect of setting up the client, and then we can call this `initializeJourney` function without arguments. I'd rather avoid this kind of pattern.

Can we move the client configuration into this `initializeJourney` function, or maybe pass in the returned result of configuration into the init function?

cerebrl (changes requested):

core/journey-client.config.ts
vatsalparikh: > Can you elaborate as to why this is outside of the `journey/` directory? Also, does this need a dedicated file, or can this be integrated into an existing one? I'm not totally against an additional file, but my first reaction is to integrate this into the `journey.store.ts`. But, I could just be missing something.

My understanding is that journey client configuration is consumed by core/journey.store.ts, by login app, and by login widget.

journey.store.ts file simply retrieves the journey client. Setting the journey client is done by login app, login widget, or e2e tests. journey.store does not need anything to be "set", journey store just needs to get the journey client.
Whereas setting the journey client is needed for login app and login widget. We can get and set the journey client in the respective places but that would disperse many things. The schema validation done during setting the journey client will be repeated if we keep setting it wherever it's needed.

We can probably move getJourneyClient to journey.store.ts but then we lose a singular place where both getting and setting happens. I thought of moving it to sdk.config.ts file but sdk.config.ts file is for synchronous stuff, so I thought a separate file is better. However, I think moving it to sdk.config.ts is the best thing, I think. This way, when we have device client or oidc client config, we can just have a single place where all those configs exist.

Let me know your thoughts.

core/journey/callbacks/_utilities/callback.utilities.ts
vatsalparikh: Added a TODO and deriving the type for now, just like I am doing it wherever we don't have a re-export

`  // TODO: PolicyParams must be re-exported by journey-client from forgerock/sdk-types, until then we derive the type like below
params: PolicyRequirement['params'];`

Will create a ticket for this and other exports

core/journey/callbacks/device-profile/device-profile.stories.js
vatsalparikh: Thanks, Justin and Ryan for the links. I've updated to using `Device` from journey-client now!

core/journey/stages/stages.story.svelte
vatsalparikh: Now that I am using Device from journey-client, https://github.com/ForgeRock/forgerock-web-login-framework/pull/468#discussion_r3171256268, `Config` is no longer needed and removed.

core/journey/callbacks/device-profile/device-profile.story.svelte
vatsalparikh: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468#discussion_r3171264035

core/journey/callbacks/device-profile/device-profile.svelte
vatsalparikh: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468#discussion_r3171256268

core/journey-client.config.ts
vatsalparikh: @cerebrl , I have moved journey client configuration into journey store. The only thing I am concerned about here is journey store is getting too big, and it's doing initialization and the journey flow both. That's why I wanted to keep a separate file. But because getJourneyClient is only used by journey store, and initialization needs to set the journey client it does make sense to move this logic to journey store. Maybe we can split up the journey store later and make it more modular. For now, I've removed the separate journey-client.config.ts file. Here is the commit: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468/changes/7a62269692dc7ca7a8e7b9b4888b7a73e2b74132

packages/login-widget/src/lib/_utilities/api.utilities.ts
vatsalparikh: Makes perfect sense. I've moved the logic to setJourneyClientConfig inside initialize along with schema validation. It's all in one file now, journey.store.ts. Here is the commit: https://github.com/ForgeRock/forgerock-web-login-framework/commit/7a62269692dc7ca7a8e7b9b4888b7a73e2b74132

core/journey/_utilities/data-analysis.utilities.ts
vatsalparikh: If we don't know the type where we make the call, we still need to keep it generic like BaseCallback.

When we extend T with BaseCallback we still can't know what exactly is T, it could be SelectIdpCallback or some other callback.

Instead, I've cast it to SelectIdpCallback from where we're calling this function. Let me know if this looks right, thanks!

https://github.com/ForgeRock/forgerock-web-login-framework/blob/080bfd4cc89b79c5a45d0d9fcb398df8942f07cf/core/journey/_utilities/data-analysis.utilities.ts#L53

https://github.com/ForgeRock/forgerock-web-login-framework/blob/080bfd4cc89b79c5a45d0d9fcb398df8942f07cf/core/journey/_utilities/data-analysis.utilities.ts#L88

cerebrl (approved):
Looks good to me.

apps/login-app/src/routes/(app)/+page.svelte
```diff
@@ -29,7 +29,11 @@
   const journeyParam = $page.url.searchParams.get('journey');
   const suspendedIdParam = $page.url.searchParams.get('suspendedId');
 
-  const journeyStore: JourneyStore = initializeJourney();
+  const journeyStore: JourneyStore = initializeJourney({
+    serverConfig: {
+      wellknown: data.wellknown,
```
SteinGabriel, line 34: I got a 500 error while testing it locally because I was missing the wellknown URL in `.env`. I think we could have more meaningful error handling instead of a blank page with 500 error message, especially because it's possible that developers running this for the first time will face the same issue.

vatsalparikh: We do have explanation in .env.example and in docs, but yeah, having a helpful error message makes sense. I've kept the 500 error as it is, but added a helpful message. Here is the commit: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468/changes/af99389a272ff818f2761bc046a3df094358a436

And here is the screenshot. Let me know if this is good, thanks!

Screenshot 2026-05-04 at 5 18 28 PM

SteinGabriel: Yeah, that's definitely more informative. I think it'd be nice if we could render this error in the app's UI, but if that's not doable, this should be informative enough. 
Thank you!

SteinGabriel (approved):

apps/login-app/src/routes/(app)/+page.svelte
vatsalparikh: Okay, I've added a proper UI when login app is not configured. Let me know if this looks good, thanks!

Screenshot 2026-05-05 at 11 58 15 AM

PR #470, opened 2026-04-21
Sdks 4612 autofill passkeys

vatsalparikh:
## JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4612

## Description
This PR adds a new feature of autofilling passkeys during authentication. This autofill option shows up only on username field in any journey that's webauthn eligible, so it is stage independent. It depends on webauthn callback and name callback, and the option is showing when both callbacks exist.

## Notes

- Right now, normal login with username and password will not work with `TEST_AutofillPasskeyWebAuthn` journey because this requires AM's 8.1 release which uses the new Page Node version 2.0. We need this Page Node for normal username / password login to work alongside autofill with passkey.
- Conditional mediation is not supported by the new Ping SDK. So this PR includes logic to test conditional mediation with browser. I am working on adding this logic to Ping SDK but it will only be available once we have an updated version of journey client published to npm. I've kept the logic minimal and separate so swapping out the logic by calling journey client should be simple.

## Recording

https://github.com/user-attachments/assets/4980c339-b93b-469d-b15f-bd08f1492da2

## How to test

- Make sure to use `http` instead of `https` while testing. Passkey option shows up only when using `http` on localhost. To do this, update the [vite.config.ts](https://github.com/ForgeRock/forgerock-web-login-framework/blob/ccc01c4b7a25462b2ddda44e5eaeb82a0ae1be0c/apps/login-app/vite.config.ts#L12) file's line 12 to only use `http` by removing `basicSsl()` like so `(mode === 'development' ? [] : []),`
- Register with webauthn by going to `TEST_WebAuthn-Registration` journey and then try authenticating by going to `TEST_AutofillPasskeyWebAuthn` journey which has the autofill option. You can follow the attached recording for reference.


core/journey/stages/_effects/webauthn.effects.ts
```diff
@@ -0,0 +1,166 @@
+/**
+ *
+ * Copyright © 2026 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ **/
+
```
SteinGabriel, line 9: I believe we should add some tests for this new functionality.

core/journey/_utilities/metadata.utilities.ts
```diff
@@ -56,6 +57,7 @@ export function buildCallbackMetadata(
         isReadyForSubmission: isCbReadyByDefault(callback),
         isSelfSubmitting: isSelfSubmitting(callback),
         isUserInputRequired: requiresUserInput(callback),
+        isPasskeyAutofillEligible: isMixedLoginWebAuthnStep(step),
```
SteinGabriel: We could just call `isMixedLoginWebAuthnStep(step)` once before the loop starts instead of doing it every time inside the callbacks.

SteinGabriel (commented):
Minor findings.

core/journey/_utilities/metadata.utilities.ts
vatsalparikh: yeah, that makes a lot of sense, interesting! Moved it at the top outside the map function, thanks!

core/journey/stages/_effects/webauthn.effects.ts
vatsalparikh: Added tests with mock, thanks!

core/journey/stages/_utilities/webauthn.utilities.ts
```diff
@@ -0,0 +1,98 @@
...
+     * which helps determine if this is a registration step
+     */
+    const hasPubKeyCredParams =
+      !!data &&
+      typeof data === 'object' &&
+      Object.prototype.hasOwnProperty.call(data, 'pubKeyCredParams');
+
+    /* Return false if this is a WebAuthn registration step. */
+    return !hasPubKeyCredParams;
+  }
```
cerebrl: Could you elaborate as to why you have this additional logic above? The below `getWebAuthnStepType` should be sufficient by itself, no?

core/journey/stages/_utilities/webauthn.utilities.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  return WebAuthn.getWebAuthnStepType(step) === WebAuthnStepType.Authentication;
+}
+
+/**
+ * @function getWebAuthnMetadataCallback - gets the WebAuthn MetadataCallback from a step
+ * @param {JourneyStep} step - The current journey step
+ * @returns {MetadataCallback | undefined} The WebAuthn metadata callback, if present
+ */
+function getWebAuthnMetadataCallback(step: JourneyStep): MetadataCallback | undefined {
```
cerebrl: I believe this too is already available as a function from the SDK: https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/journey-client/src/lib/webauthn/webauthn.ts#L258

core/journey/stages/_utilities/webauthn.utilities.ts
```diff
@@ -0,0 +1,98 @@
...
+      return Object.prototype.hasOwnProperty.call(data, 'relyingPartyId');
+    });
+}
+
+/**
+ * @function hasWebAuthnOutcomeCallback - checks if the step includes the WebAuthn outcome callback
+ * @param {JourneyStep} step - The current journey step
+ * @returns {boolean} True if a HiddenValueCallback with id "webAuthnOutcome" exists
+ */
+function hasWebAuthnOutcomeCallback(step: JourneyStep): boolean {
```
cerebrl: This too: https://github.com/ForgeRock/ping-javascript-sdk/blob/main/packages/journey-client/src/lib/webauthn/webauthn.ts#L272

core/journey/journey.svelte
```diff
@@ -50,8 +59,8 @@
     }
   }
 
-  afterUpdate(() => {
-    alertNeedsFocus = $journeyStore && !$journeyStore.successful;
```
cerebrl, line 54: Can you elaborate on why this was removed?

core/journey/stages/_effects/webauthn.effects.ts
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * @function authenticateWebAuthnStep - authenticates a WebAuthn step, optionally using conditional mediation
+ * @param {JourneyStep} step - The WebAuthn journey step
+ * @param {boolean} [useConditionalMediation] - True to use conditional mediation (passkey autofill)
+ * @param {AbortController} [abortController] - Abort controller required for conditional mediation
+ * @returns {Promise<JourneyStep>} The same step with the WebAuthn outcome written to the hidden callback
+ * @throws {Error} If conditional mediation is requested without an AbortController
+ * @throws {Error} If the step does not contain the expected WebAuthn callbacks
+ */
+export async function authenticateWebAuthnStep(
```
cerebrl: Should I assume this is being used because we *_don't_* have Conditional UI support in the new Ping SDK? So, if the SDK supports Conditional UI, then this is no longer needed?

core/journey/stages/_effects/webauthn.effects.test.ts
cerebrl, line 1: The complexity of all of this mocking is usually a code smell for me. If this level of mocking is needed, it usually means the type of testing is ill-suited for what is being tested. This is essentially trying to test browser level functionality within a unit test, yeah? So, I'd try to move this to a testing type that has a browser, like `e2e`.

For unit tests, the only thing that should be mocked is the data. For integration testing, you might mock data and a function or two. This test is mocking out a lot more than just a function or two, so I think it's too much for this test type.

cerebrl (changes requested):
Comments from an initial scan.

core/journey/stages/_utilities/webauthn.utilities.ts
vatsalparikh: That was an oversight on my part, sorry. Will keep this in mind to ensure it never happens again.

I've updated the webauthn.utilities file, it's pretty lean now with only 1 function and using the available Webauthn helpers from journey-client.

core/journey/stages/_utilities/webauthn.utilities.ts
vatsalparikh: That was an oversight on my part, sorry. Will keep this in mind to ensure it never happens again.

I've updated the webauthn.utilities file, it's pretty lean now with only 1 function and using the available Webauthn helpers from journey-client.

core/journey/stages/_utilities/webauthn.utilities.ts
vatsalparikh: That was an oversight on my part, sorry. Will keep this in mind to ensure it never happens again.

I've updated the webauthn.utilities file, it's pretty lean now with only 1 function and using the available Webauthn helpers from journey-client.

core/journey/stages/_effects/webauthn.effects.ts
vatsalparikh: Yes, both functions `authenticateWebAuthnAutofill` and `isConditionalMediationSupported` are currently checking for conditional mediation support with the browser and also calling navigate.credentials with mediation and signal. I've added a TODO to update `authenticateWebAuthnAutofill` and remove `isConditionalMediationSupported` when journey-client logic is implemented and available in npm.

core/journey/stages/_effects/webauthn.effects.test.ts
vatsalparikh: I've reduced the mock as much as possible in this file. We do need these tests because playwright e2e tests cannot test autofill passkey. This is because passkey dropdown is a browser feature not available to DOM. 

I looked at CDP virtual authenticators but could not find any autofill for passkeys. CDP virtual authenticators have support for normal autofill like address and forms, but not for passkey. 

So the closest we can get to e2e test here is by integration test here where we mock the journey but use real webauthn functions. I already have e2e tests under widget modal and inline test, but the extent of those is only checking whether the autofill attribute is visible. Unable to complete the flow by clicking on that passkey dropdown. So combining those e2e with this integration test is the best for autofill passkey.

Let me know if that looks right after reducing the mocking in webauth.effects.test.ts file

core/journey/journey.svelte
vatsalparikh: `alertNeedsFocus` was simply moved to a reactive block instead of keeping it in afterUpdate: https://github.com/ForgeRock/forgerock-web-login-framework/pull/470/changes#diff-e0c4cf7e1aa26f05a102129c95d7722dd05bec2e068c9da1032d9413f7e440d7L33

journey.svelte is a very high traffic file. We only need alertNeedsFocus when the journey is not successful rather than after every update, so keeping it in afterUpdate is wasteful and unnecessary. Besides, `alertNeedsFocus` is a derived state, so it belongs to the reactive block.

core/journey/stages/_effects/webauthn.effects.test.ts
cerebrl: Okay, you make a good argument. Thanks for the response! Can you essentially add this as a code comment in the file, so the next person that comes along understands the need for this test?

core/journey/journey.svelte
cerebrl: Hmm. I feel like there was a reason for this being in an `afterUpdate`, but I can't think of it right now. This is directly for the purpose of accessibility, so I want to make sure to get this right. If in your manual testing, you have found that moving it to the reactive block results in the intended result, I'm okay with it.

I'm just very sensitive to things related to accessibility.

cerebrl (approved):

core/journey/journey.svelte
vatsalparikh: If it pertains to accessibility, I can imagine the side effects not running when they should if the variable is in a derived state.

While I did not find issues with autofill passkey, this is the journey.svelte file where there are so many other journeys, steps, and callbacks which show this focus, so your concern is valid.

I moved back alertNeedsFocus under afterUpdate.

core/journey/stages/_effects/webauthn.effects.test.ts
vatsalparikh: Done, added a comment, thanks!

SteinGabriel (approved):
Changes look good. Tested in the browser and it worked as expected.

PR #471, opened 2026-04-22
chore: update-deps

ryanbas21:
update zod, xss, remove axios


.changeset/upgrade-zod-v4-and-xss.md
```diff
@@ -0,0 +1,16 @@
+---
+'@forgerock/login-widget': minor
+---
+
+Upgrade `zod` from `3.22.4` to `^4.3.6` and `xss` from `1.0.14` to `^1.0.15`.
+
+- Migrate `core/sdk.config.ts` config schema to Zod v4: replace the removed
+  `z.function().args().returns()` schema form with `z.custom<Fn>()` for the
+  `callbackFactory`, `middleware`, and `tokenStore` function validators;
+  switch `z.nativeEnum(CallbackType)` to `z.enum(CallbackType)`; collapse
```
SteinGabriel: It looks like `z.nativeEnum` switched to `z.custom` instead of `z.enum` in `sdk.config.ts`.

SteinGabriel (approved):
Other than a small find in the changeset and a couple of out-of-date copyright dates, it looks good to be merged.
Out-of-date copyright: 
- `sdk.config.ts`
- `i18n.utilities.ts`

.changeset/upgrade-zod-v4-and-xss.md
ryanbas21: Fixed — the changeset incorrectly described this as a `z.nativeEnum → z.enum` switch. What actually happened is the entire `callbackFactory` validator (which included `z.nativeEnum(CallbackType)`) was collapsed into `z.custom<CallbackFactoryFn>(isFunction)`, so there is no `z.enum` usage. Updated the wording to reflect that.

PR #473, opened 2026-04-29
fix(import-order): fix import order for all files by defining an esli…

vatsalparikh:
This PR fixes the import orders for all files in login framework.

Justin provided feedback on import order in the journey client migration PR: https://github.com/ForgeRock/forgerock-web-login-framework/pull/468#discussion_r3149204326

Based on his comment on another PR about the expected import order: https://github.com/ForgeRock/ping-javascript-sdk/pull/541/changes#r2956779035 I am creating this new PR which specifies the exact import order using an eslint rule.

Here is the import order rule:

// import any proper modules first
import { StepType } from '@forgerock/sdk-types';

// import any file-based modules second
import { journey } from './client.store.js';

// import any types from proper modules third
import type { GenericError } from '@forgerock/sdk-types';

// import any types from file-based modules last
import type { SomeType } from './some-file.js';

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,9 +1,16 @@
 {
+  "plugins": ["prettier-plugin-svelte"],
   "useTabs": false,
   "singleQuote": true,
   "trailingComma": "all",
   "printWidth": 100,
   "overrides": [
+    {
+      "files": "*.svelte",
+      "options": {
+        "parser": "svelte"
+      }
+    },
     {
       "files": "*.json",
       "options": {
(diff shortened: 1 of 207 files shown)


ryanbas21 (approved):
I believe all the changes here are eslint/automatic so i just reviewed the eslint config changes.

if i'm right as long as CI is happy 👍

vatsalparikh:
> I believe all the changes here are eslint/automatic so i just reviewed the eslint config changes.
> 
> if i'm right as long as CI is happy 👍

Yes, I verified each file change yesterday, and it strictly follows all the rules to the T. We just need to review the import order in eslint, and see if it's not too strict. Either way, we can update it later if needed!

PR #478, opened 2026-04-30
SDKS-2810: Add Invisible reCAPTCHA, hCaptcha, and Enterprise

SteinGabriel:
## Summary

https://pingidentity.atlassian.net/browse/SDKS-2810

Adds invisible reCAPTCHA v2 and invisible hCaptcha support to the login widget via a new `configuration({ captcha: { mode: 'invisible' } })` option. Also adds a dedicated `ReCaptchaEnterpriseCallback` handler for AM journeys using the Enterprise CAPTCHA node. The widget now auto-injects CAPTCHA scripts — no manual `<script>` tag required from consumers for any variant.

---

### `core/captcha.config.ts` 

- Zod schema (`captchaConfigSchema`) for the `captcha` config option — validates `mode: 'visible' | 'invisible'`. Strict so unknown keys throw at parse time.

### `core/journey/_utilities/captcha.effects.ts` 

- `resolveGrecaptcha()` — prefers `window.grecaptcha.enterprise`, falls back to `window.grecaptcha`. Transparent to consumers regardless of which script they loaded (classic `api.js` or `enterprise.js`).
- `loadCaptchaScript({ src, provider })` — injects a `<script>` tag and resolves when the provider API is ready. No-ops if the provider is already on `window` (pre-loaded or test stub). Race-safe: `settled` flag, `error` listener, TOCTOU double-check.
- `renderCaptcha(...)` — visible v2 render for both Google and hCaptcha.
- `renderCaptchaInvisible(...)` — invisible v2 render + execute for both providers.

### `core/journey/stages/_effects/captcha.effects.ts` 

- Re-exports `renderCaptcha`, `renderCaptchaInvisible`, `resolveGrecaptcha` from `_utilities/captcha.effects.ts`. Preserves the layer boundary so `stages/` consumers keep their existing import path.

### `core/journey/callbacks/recaptcha-enterprise/` 

- `recaptcha-enterprise.svelte` — handles `ReCaptchaEnterpriseCallback`. Visible checkbox or score-based invisible flow driven by `captchaMode`. Calls `setAction()`, `setResult()`, and `setClientError()` on execute. Inline `<Alert type="error">` on failure.
- `recaptcha-enterprise.effects.ts` — `renderEnterpriseCaptcha`, `executeEnterpriseCaptcha`, `loadEnterpriseScript`. `loadEnterpriseScript` appends `?render=SITE_KEY` for invisible/score-based keys (required for `execute()`).
- `recaptcha-enterprise.utilities.ts` — `buildEnterpriseScriptSrc` pure helper.

### `core/journey/callbacks/recaptcha/recaptcha.svelte`

- Extended for invisible v2: awaits `loadCaptchaScript`, then calls `renderCaptchaInvisible` on mount for both Google and hCaptcha.
- `captchaMode` driven by `callbackMetadata.initOptions.mode` (threaded via `buildCallbackMetadata` — no store).
- `CaptchaMode` typed declaration prevents reactive widening to `string`.
- Script auto-injection: `onMount` awaits `loadCaptchaScript` before render; v3 path also self-injects `api.js`.

### `core/journey/_utilities/metadata.utilities.ts`

- `buildCallbackMetadata` accepts a new `initializationOptions` param. For `ReCaptchaCallback` and `ReCaptchaEnterpriseCallback`, merges `captcha` and `recaptchaAction` from `initializationOptions` into `initOptions` on the returned metadata object. Both callback components read mode and action from `callbackMetadata.initOptions` — no store subscriptions.

### `core/journey/_utilities/callback-mapper.svelte`

- Registered `CallbackType.ReCaptchaEnterpriseCallback` → `recaptcha-enterprise.svelte`.

### `core/journey/journey.interfaces.ts`

- `CaptchaMode = 'visible' | 'invisible'` exported type — shared across both callback components.

### `packages/login-widget/src/lib/interfaces.ts` + `api.utilities.ts`

- `captcha?: { mode?: CaptchaMode }` option on `WidgetConfigOptions`. Validated via `captchaConfigSchema.parse()` at both `initialize` and `set` call sites.

### `packages/login-widget/README.md`

- CAPTCHA Configuration section updated: accurate script auto-injection behavior, `'visible' | 'invisible'` mode values, Enterprise backward-compatibility note.

---

### `captchaMode` threading (no store)

`captcha.store.ts` was removed during review. `captchaMode` flows through `buildCallbackMetadata`'s `initializationOptions` param → `CallbackMetadata.initOptions.mode` → component prop. This keeps the value co-located with the callback that uses it, consistent with the WebAuthn autofill pattern.

### Script injection

The widget now owns CAPTCHA script injection for all variants. `loadCaptchaScript` early-exits if the provider API is already on `window`, so pre-loading in the host page still works. Enterprise invisible keys require `?render=SITE_KEY` in the script URL for `execute()` — `buildEnterpriseScriptSrc` handles this automatically.

### Layer boundary

Callbacks import shared effects from `core/journey/_utilities/captcha.effects.ts`. `stages/` code re-exports from there via `stages/_effects/captcha.effects.ts`. Callbacks never import from `stages/`.

---

## Tests

- **Unit (156 pass):**
  - `captcha.effects.test.ts` — 7 tests (`resolveGrecaptcha`, `loadCaptchaScript`: new inject, reuse existing, grecaptcha.ready, onerror reject)
  - `stages/_effects/captcha.effects.test.ts` — 12 tests (`renderCaptcha`, `renderCaptchaInvisible`, Enterprise namespace, classic fallback)
  - `recaptcha-enterprise.effects.test.ts` — 9 tests (`renderEnterpriseCaptcha`, `executeEnterpriseCaptcha`, null-grc guard)
  - `recaptcha-enterprise.utilities.test.ts` — 2 tests (`buildEnterpriseScriptSrc`)
  - `recaptcha.utilities.test.ts` — 2 tests (`checkForHCaptcha`)
  - `metadata.utilities.test.ts` — 10 tests (6 new `initializationOptions` coverage)
- **Storybook stories:** `VisibleGoogle`, `VisibleHCaptcha`, `InvisibleGoogle`, `InvisibleError` + `VisibleEnterprise`, `InvisibleEnterprise`, `InvisibleEnterpriseError`
- **Playwright E2E (6 tests — modal + inline suites):**
  - Classic visible token submit
  - Enterprise visible render check
  - Enterprise invisible execute + action + token submit
  - All use `route.fulfill` AM mocking — no live AM dependency

---

### 1. Unit tests

```shell
pnpm --filter @forgerock/login-widget exec vitest run 'captcha'
```

All 156 tests should pass.

### 2. Storybook

```shell
pnpm storybook
```

Navigate to **Callbacks/ReCaptcha** and **Callbacks/ReCaptchaEnterprise** — verify all story variants render without errors.

### 3. E2E tests

```shell
pnpm build:app
pnpm --filter @forgerock/login-widget-e2e exec playwright install chromium
pnpm ci:e2e -- tests/widget/modal/widget-modal.recaptcha.test.js
pnpm ci:e2e -- tests/widget/inline/widget-inline.recaptcha.test.js
```

#### Setting up a CAPTCHA node in an AM journey

**Classic reCAPTCHA v2 (visible or invisible):**

1. In AM admin console, open or create a journey.
2. Add a **CAPTCHA** node (`Authentication Nodes → CAPTCHA`).
3. Set **CAPTCHA Site Key** to your Google reCAPTCHA v2 site key (registered for your domain + `localhost`).
4. Set **CAPTCHA Secret Key** to the corresponding secret key — AM uses this for server-side verification.
5. For **hCaptcha**: set the site key and secret key from your hCaptcha dashboard. The node sets `captchaDivClass: h-captcha` automatically for hCaptcha keys.
6. Wire the node between `Start` and your success/failure outcomes.

**Invisible mode:** No AM config change needed. Pass `configuration({ captcha: { mode: 'invisible' } })` in the widget consumer.

**reCAPTCHA Enterprise:**

1. Add a **reCAPTCHA Enterprise** node (`Authentication Nodes → reCAPTCHA Enterprise`).
2. Set the **Enterprise Site Key** (from Google Cloud Console, reCAPTCHA Enterprise product — must be a v2 checkbox key for visible mode or a score-based key for invisible/score mode).
3. The node sends `ReCaptchaEnterpriseCallback` — the widget handles this automatically via the new component.
4. For score-based (invisible) keys: pass `configuration({ captcha: { mode: 'invisible' } })`.

#### Verifying the integration

**Visible mode (v2 checkbox):**
- The CAPTCHA checkbox widget renders inside the login form.
- Click the checkbox — it should show the green checkmark (or trigger an image challenge).
- Click **Sign In** — the journey should advance. If AM server-side verification fails, the journey returns to the CAPTCHA step.

**Invisible mode (v2 invisible / hCaptcha invisible):**
- No visible widget appears in the form.
- On mount, the widget executes the CAPTCHA automatically in the background.
- If a challenge is triggered by Google/hCaptcha (low-confidence score), a challenge popup appears.
- Once resolved, **Sign In** proceeds automatically — the token is already set.
- On failure or expiry, an inline error alert appears: *"CAPTCHA verification failed. Please try again."*

**Enterprise invisible (score-based v3-style):**
- No visible widget.
- `execute()` fires on mount with the configured `recaptchaAction` (e.g. `'LOGIN'`).
- Token, action, and any client error are submitted to AM.
- Verify in AM audit logs that the assessment score is present in the session.

**Verifying token submission via browser DevTools:**
1. Open DevTools → Network tab.
2. Trigger the CAPTCHA step.
3. Look for the AM journey `authenticate` POST request.
4. In the request body, find the `callbacks` array. The `ReCaptchaCallback` or `ReCaptchaEnterpriseCallback` input should have a non-empty `value` for `IDToken<n>token`.
5. For Enterprise, also verify `IDToken<n>action` is set to your configured action string.

**Verifying Enterprise backward compatibility:**
- Load `enterprise.js` — confirm `resolveGrecaptcha()` resolves to `window.grecaptcha.enterprise`.
- Load `api.js` instead — confirm it falls back to `window.grecaptcha` without errors.
- Both flows should complete the journey successfully.


vatsalparikh (changes requested):
I see that Gabriel has added callbacks and stages using the old legacy forgerock sdk, so this will stop working completely as soon as my journey client changes are merged to main. I suggest we use the new journey client here, because my journey client PR https://github.com/ForgeRock/forgerock-web-login-framework/pull/468 will soon be merged to main.

apps/login-app/src/app.html
```diff
@@ -19,6 +19,11 @@
         document.body.classList.add('tw_dark');
       }
     </script>
+    <!--
+      For CAPTCHA-enabled journeys, load the reCAPTCHA Enterprise script:
```
vatsalparikh: This comment seems misplaced? Was this comment intended for layout.svelte file linked below?

https://github.com/ForgeRock/forgerock-web-login-framework/blob/48f1ac633b9017e28dc0b56f43f34eb6db6c1645/apps/login-app/src/routes/(app)/%2Blayout.svelte#L27

apps/login-app/src/routes/(app)/+page.svelte
```diff
@@ -40,6 +41,7 @@
    * Sets up locale store with appropriate content
    */
   initializeContent(data.content);
+  initializeCaptcha({ mode: 'invisible' });
```
vatsalparikh: Is captchaMode query param used for hCaptcha?

We already have recaptchaAction option which is set in journey store when we start the journey. We should either combine recaptchaAction with captchaMode or add support for both in journey.store as well. Right now, I only see recaptcaAction being supported in journey store.

Or, is captchaMode only needed in query param and never needs to be passed to be stored in journey store? I believe we store recaptchaAction in journey store so the user does not have to keep performing captcha check at every step of the journey. So we should have that support for hCaptcha as well.

Either way, I think we need to look deeper into supporting both in journey store or know why we don't them both in journey store.

core/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -46,31 +52,47 @@
   const recaptchaClass: string =
     callback?.getOutputByName<string>('captchaDivClass', 'h-captcha') ?? 'h-captcha';
 
+  const captchaProvider: 'hcaptcha' | 'grecaptcha' =
```
vatsalparikh, line 51: We are handling both hCaptcha and recaptcha in this file and in the folder in general, so we should rename all files under and along with `recaptcha` folder to be `captcha` instead

core/journey/stages/_utilities/recaptcha.utilities.ts
```diff
@@ -10,40 +10,89 @@
...
  */
+
+/**
+ * Resolves the active reCAPTCHA namespace. Prefers `grecaptcha.enterprise`
+ * (loaded by enterprise.js) and falls back to the classic `grecaptcha` global
+ * (loaded by api.js or auto-migrated keys). This makes all call sites
+ * transparent to which script the consumer loaded.
+ */
+export function resolveGrecaptcha(): ReCaptchaV2.ReCaptcha {
+  const g = window.grecaptcha as ReCaptchaV2.ReCaptcha & { enterprise?: ReCaptchaV2.ReCaptcha };
```
vatsalparikh: I think we should avoid single letter variable names

core/journey/stages/_utilities/recaptcha.utilities.ts
```diff
@@ -10,40 +10,89 @@
...
+        callback: window.frHandleCaptcha,
+        'expired-callback': window.frHandleExpiredCallback,
+        'error-callback': window.frHandleCaptchaInvisibleError,
+      });
+      grc.execute(widgetId);
+    });
+  }
+}
+
 export function handleCaptchaError(callback: ReCaptchaCallback) {
```
vatsalparikh: Now that we have a dedicated recaptcha.utilities.test.ts file we should add unit tests for `handleCaptchaError` function as well.

e2e/tests/widget/modal/widget-modal.recaptcha.test.js
```diff
@@ -0,0 +1,184 @@
...
+        ],
+        input: [{ name: 'IDToken1', value: '' }],
+        _id: 0,
+      },
+    ],
+    (body) => submitted.push(body),
+  );
+
+  const { clickButton, navigate } = asyncEvents(page);
+  await navigate('widget/modal?journey=TEST_Login');
```
vatsalparikh, line 115: I think we should also add e2e tests for widget inline file if we have tests for widget modal, unless recaptcha is only applicable for widget modal cases.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
```diff
@@ -0,0 +1,118 @@
...
+  import { journeyStore } from '$journey/journey.store';
+
+  import type { SelfSubmitFunction, StepMetadata } from '$journey/journey.interfaces';
+  import type { styleSchema } from '$core/style.store';
+  import type { Maybe } from '$core/interfaces';
+  import { renderCaptcha } from '$journey/stages/_utilities/recaptcha.utilities';
+  import { executeEnterpriseCaptcha, detectEnterpriseProvider } from './recaptcha-enterprise.utilities';
+  import { captchaStore } from '$core/captcha.store';
+  import Alert from '$components/primitives/alert/alert.svelte';
+  import { interpolate } from '$core/_utilities/i18n.utilities';
```
cerebrl: I'd like to reorder these imports to the below pattern with a blank lines separating the groups. Imports should be alphabetized according to the imported filename: 

```
// import external runtime dependencies

// import external type dependencies

// import internal runtime dependencies

// import internal type dependencies

// rest of code ...
```

e2e/tests/widget/modal/widget-modal.recaptcha.test.js
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ **/
+
+import { expect, test } from '@playwright/test';
+import { asyncEvents } from '../../utilities/async-events.js';
+
+async function stubGrecaptchaClassic(page) {
```
vatsalparikh, line 18: Is there a reason for subbing here in this e2e test? Is it because we don't have keys for classic recaptcha?

packages/login-widget/src/lib/_utilities/api.utilities.ts
```diff
@@ -107,6 +108,7 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
     initializeJourneys(options?.journeys);
     initializeLinks(options?.links);
     initializeStyle(options?.style);
+    initializeCaptcha(options?.captcha);
```
vatsalparikh: Should we set options?.captcha somewhere in the login widget like we do for recaptchaAction? https://github.com/ForgeRock/forgerock-web-login-framework/blob/48f1ac633b9017e28dc0b56f43f34eb6db6c1645/packages/login-widget/src/lib/interfaces.ts#L38

So it would be something like 

export interface JourneyOptionsStart {
  forgerock?: StepOptions;
  journey?: string;
  resumeUrl?: string; // current URL if resuming a journey/tree
  recaptchaAction?: string;
  captcha?: string
}

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
```diff
@@ -0,0 +1,118 @@
...
+<script lang="ts">
+  import type { ReCaptchaEnterpriseCallback } from '@forgerock/javascript-sdk';
+  import type { z } from 'zod';
+  import { onMount } from 'svelte';
+  import { journeyStore } from '$journey/journey.store';
+
+  import type { SelfSubmitFunction, StepMetadata } from '$journey/journey.interfaces';
+  import type { styleSchema } from '$core/style.store';
+  import type { Maybe } from '$core/interfaces';
+  import { renderCaptcha } from '$journey/stages/_utilities/recaptcha.utilities';
```
cerebrl: I'd like to avoid this pattern of a callback component relying on a stage utility. To me, a stage-related file can pull in a callback-related file, but the reverse is an anti-pattern. This is a part of the hierarchical design of the system.

vatsalparikh (changes requested):
Initial review comments

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
```diff
@@ -0,0 +1,118 @@
...
+        callback?.setResult('');
+        renderCaptcha({ nameOfCaptcha: captchaProvider, siteKey, elementId: 'fr-recaptcha-enterprise' });
+      };
+      renderCaptcha({ nameOfCaptcha: captchaProvider, siteKey, elementId: 'fr-recaptcha-enterprise' });
+    }
+  });
+
+  journeyStore.subscribe((value) => {
+    recaptchaAction = value?.recaptchaAction ?? '';
+  });
```
cerebrl: This concerns me. I'd really like to find a way to do this that doesn't involve pulling in the entire Journey Store into a callback component.

core/journey/stages/_utilities/recaptcha.utilities.ts
```diff
@@ -10,40 +10,89 @@
...
   siteKey,
+  elementId = 'fr-recaptcha',
 }: {
   nameOfCaptcha: 'hcaptcha' | 'grecaptcha';
   siteKey: string;
+  elementId?: string;
 }) {
   if (nameOfCaptcha === 'hcaptcha' && window.hcaptcha) {
-    return window.hcaptcha.render('fr-recaptcha', {
+    return window.hcaptcha.render(elementId, {
```
cerebrl: This deviates from our conventions of utility functions being pure, stateless functions. Typically, anything that involves DOM or Window use is in a Svelte component. But, if something needs to access external systems, we would call it an "effect" (not to be confused with Effect, the library). 

With that said, I see that this file already has `window` use, so we can leave it. But, I do want to stress that  we should move away from this as it should be considered an anti-pattern.

cerebrl (changes requested):
I'd really like for us to reconsider some architectural choices here. If you want, we can jump on a call tomorrow and discuss it.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
vatsalparikh: I already have the import order PR approved, and we should be able to handle it automatically. As soon as I merge journey client, I will merge import order PR as well.

https://github.com/ForgeRock/forgerock-web-login-framework/pull/473

Just wanted to put it out here so Gabriel doesn't spend time fixing it.

core/journey/stages/_utilities/recaptcha.utilities.ts
SteinGabriel: Agreed. We can tackle this alongside the stage import in `recaptcha.svelte` callback file in a new ticket.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
SteinGabriel: Exported a `recaptchaActionStore` derived store from `journey.store.ts` that exposes only `recaptchaAction`. Both captcha callback components now subscribe to that instead of the full store.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
SteinGabriel: For sure, that makes sense. I moved `resolveGrecaptcha` inline into the enterprise utilities file and added a `renderEnterpriseCaptcha` helper. No more stage imports from this callback file.
 
Note: `recaptcha.svelte` still imports `handleCaptchaError`, `handleCaptchaToken`, `renderCaptcha`, `renderCaptchaInvisible`, and `resolveGrecaptcha` from the stage utilities file. That's a pre-existing pattern. Fully decoupling it would require either duplicating those helpers or extracting them to a shared non-stage utilities file, which feels like a larger refactor. Should we include that in this PR or create a tech debt ticket instead?

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
SteinGabriel: Updated the SDK import to `journey-clients` and applied the import ordering convention. Since this is a new file, it was not updated automatically after journey-client PR was merged.

packages/login-widget/src/lib/_utilities/api.utilities.ts
SteinGabriel: `captchaMode` is rendering config that controls how the captcha is rendered, not AM data that changes per journey step like `recaptchaAction`. It already lives on `configuration()` which seems to be the right place for widget-level render config. 
Let me know what you think. Thanks!

e2e/tests/widget/modal/widget-modal.recaptcha.test.js
SteinGabriel: Yeah, two reasons actually: no real recaptcha keys available in CI as you mentioned, and E2E tests shouldn't depend on third-party services, since we're testing widget behavior, not Google's API. Also noticed the real `enterprise.js` script was loading in tests and overwriting the stub, which was causing the classic visible test to flake.

e2e/tests/widget/modal/widget-modal.recaptcha.test.js
SteinGabriel: Good call. Added `widget-inline.recaptcha.test.js` with the same 3 test cases as the modal suite. Also wired `?captchaMode=` param support into the inline e2e page.

core/journey/stages/_utilities/recaptcha.utilities.ts
SteinGabriel: For sure. I added 3 tests covering the hCaptcha branch, grecaptcha branch, and the missing-class default case. Thanks!

core/journey/stages/_utilities/recaptcha.utilities.ts
SteinGabriel: Yeah, you're right. Renamed  `g` to `grecaptcha`.

core/journey/callbacks/recaptcha/recaptcha.svelte
SteinGabriel: Fair point. I think we should add this to a new tech debt ticket, along with a few other comments from this PR, to avoid adding more changes into an already large diff.

apps/login-app/src/routes/(app)/+page.svelte
SteinGabriel: `captchaMode` and `recaptchaAction` intentionally live in separate stores: `captchaMode` is a widget-level config that drives render behavior for both google and hCaptcha across all steps, while `recaptchaAction` is a per-journey option in the journey store. They serve different purposes and I don't think they need merging. 
I've also fixed the hardcoded `mode: 'invisible'` in `page.svelte`. It now reads `?captchaMode=` from the URL and defaults to `'normal'`. 
Thanks!

apps/login-app/src/app.html
SteinGabriel: Good catch. Moved the comment to `layout.svelte` where it belongs.
Thank you!

core/captcha.store.ts
cerebrl, line 1: This store feels like it's a lot for just a single string value, essentially. I'm also a bit unsure about it living at the root of the `core` directory. 

I'd like to take advantage of the Callback Metadata structure we have for this value. This value seems to only be relevant for the Enterprise Callback, and nothing else. So, it's prominence in this project is quite asymmetrical to its actual use.

Can we pass this value into the system using the existing code and attach it to the appropriate callback using the Callback Metadata structure. I know @vatsalparikh used this very pattern for his WebAuthn Autofill work, which I really liked. Can we think on this a bit more this?

cerebrl (changes requested):

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
cerebrl: Sorry, I wrote this comment before I saw this one: https://github.com/ForgeRock/forgerock-web-login-framework/pull/478#discussion_r3191835270

apps/login-app/src/routes/(app)/+page.svelte
vatsalparikh: Thanks for clarifying!

packages/login-widget/src/lib/_utilities/api.utilities.ts
vatsalparikh: Sounds good, thanks for the explanation

core/journey/callbacks/recaptcha/recaptcha.mock.ts
```diff
@@ -0,0 +1,66 @@
+/**
+ *
+ * Copyright © 2026 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ **/
+
+import { CallbackType } from '@forgerock/javascript-sdk';
```
vatsalparikh: I see you updated the repo to journey client, which is great!

We just need to clean up some of the remaining forgerock/javascript-sdk imports in 
- core/journey/callbacks/recaptcha/recaptcha.mock.ts
- core/journey/callbacks/recaptcha/recaptcha.stories.js
- core/journey/callbacks/recaptcha/recaptcha.story.svelte
- core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.mock.ts
- core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.stories.js
- core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.story.svelte

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
vatsalparikh: It should apply to new or existing files just the same. Dunno why that did not work for you. I reshuffled a few imports in [core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte](https://github.com/ForgeRock/forgerock-web-login-framework/pull/478/files#diff-a0ef90f0481ee958b0a4b69089aa994133ba1de9b1ce20352b89106f2c587653) and it put them back in the import order.

packages/login-widget/package.json
```diff
@@ -54,11 +54,10 @@
     "@sveltejs/package": "^2.5.7",
     "@sveltejs/vite-plugin-svelte": "^4.0.4",
     "@types/qrcode": "1.5.5",
-    "@vitest/coverage-v8": "^3.2.4",
```
vatsalparikh, line 57: Is there a reason to remove vitest coverage or was it accidental?

vatsalparikh (changes requested):
Overall, looking good now, the codebase looks good. 

Justin already mentioned comments about using callback metadata (or similar) instead of captcha store and putting `window` in a separate file so those utility files should become leaner, and any effectful code should be under effects file for both enterprise and normal recaptcha.

I tested recaptcha enterprise with login journey. Gabriel helped configuring a new recatpcha enterprise, and I was able to solve the recaptcha but login failed. It's possible that the new config was missing something.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
SteinGabriel: Confirmed, `recaptcha-enterprise.svelte` import order is correct and ESLint passes clean. The auto-fix must not have run on this new file initially because the import ordering PR was not merged yet at that point. The imports must've got ordered when I updated this branch from main after that PR was merged.

core/journey/callbacks/recaptcha/recaptcha.mock.ts
SteinGabriel: Done, replaced `CallbackType` and `FRStep` from `@forgerock/javascript-sdk` with `callbackType` and `createJourneyStep` from `@forgerock/journey-client` across all 6 files. Thanks for catching those!

By the way, I just realized that the custom callback and stage components generated by the CLI tool are created using the legacy SDK instead of journey-client, and also need to be updated. 
I'm not sure whether this should be in the scope of this ticket or addressed in a separate ticket.

vatsalparikh: Ah, the CLI tool!

I did update the demo custom components in journey client migration ticket, so that's good. 

I think CLI tool can be done separately since it is experimental.

core/captcha.store.ts
SteinGabriel: Looked into this more carefully. I think `captchaMode` can't fit the Callback Metadata pattern that returns `derived` from `buildCallbackMetadata`(the autofill path @vatsalparikh used), which computes values from the step/callback structure in widget code. `isPasskeyAutofillEligible` works because the step structure itself signals eligibility (`NameCallback` + WebAuthn metadata + outcome callback all present). `captchaMode` has no equivalent signal in the callback payload. Neither `ReCaptchaCallback` nor `ReCaptchaEnterpriseCallback` expose an output field for invisible mode. 

That said, the store for one value concern is fair. Happy to fold `captchaMode` into the existing `journey/config.store` (removes the separate file, no consumer API change) or move `captcha.store.ts` from `core/` root into `core/journey/` so its scope feels more proportional. Let me know which direction you'd prefer. Thank you!

packages/login-widget/package.json
SteinGabriel: Not really, that was accidental. 
Restored `@vitest/coverage-v8` to `devDependencies`.
Thanks for catching that!

core/captcha.store.ts
cerebrl: Hey Gabriel, thanks for the thoughtful reply. I agree with all your points, but want to add that the `buildCallbackMetadata` is not exclusively for deriving data. It does do that, BUT its original intention is to accumulate a data structure that comes from non-callback sources to help inform how the callback should be rendered/managed. It doesn't have to exclusively be derived from callbacks themselves.

This broader intention is why `derived` is a property of the object, not the object itself. There's nothing stopping us from adding a different property, say `initConfig` or `initOptions` that could contain this `mode: 'invisible'` key-value pair. So, in other words:

```
// Pseudocode
export function buildCallbackMetadata(
  step: FRStep,
  checkValidation: (callback: FRCallback) => boolean,
  stageJson?: Record<string, unknown> | null,
  initializationOptions?: Record<string, unknown> | null, <-- it just needs 
) {
  const callbackCount: Record<string, number> = {};

  return step?.callbacks.map((callback, idx) => {
    let initOptions;

    // ...

    if (callback.type === <captcha callback>) {
      initOptions = initializationOptions.captcha;
    }

    return {
      derived: {
        // ...
      },
      ...(initOptions && { initOptions }), <-- new property
      idx,
      // Only use the `platform` prop if there's metadata to add
      ...(stageCbMetadata && {
        platform: {
          ...stageCbMetadata,
        },
      }),
    };
  });
}
```

SteinGabriel: Got it, thanks for clarifying. I was reading `buildCallbackMetadata` as a mapping of callback metadata coming from AM, not as local metadata config. It makes perfect sense to me now, agree it's the right place for captcha init settings.

That said, ``captcha.store.ts`` has been deleted and replaced it with the `initOptions` pattern from your pseudocode. `buildCallbackMetadata` now accepts an `initializationOptions` param, for `ReCaptchaCallback` and `ReCaptchaEnterpriseCallback`, it attaches `initOptions = initializationOptions.captcha` to the returned metadata object. Both callback components receive `callbackMetadata` as a prop and read
  mode from `callbackMetadata?.initOptions?.mode`. The captcha config option on `configuration()` / `WidgetConfigOptions` is preserved — it's passed into `initializeJourney()` as `{ captcha }` and threaded through to `buildCallbackMetadata`.

apps/login-app/src/routes/(app)/+layout.svelte
```diff
@@ -23,10 +23,11 @@
...
   <meta name="viewport" content="width=device-width, initial-scale=1" />
-  <script
-    src="https://www.google.com/recaptcha/api.js?render=6LdIqXMoAAAAAP4APBlw7_5WDeMTlAAQJf42rPWz"
-    async
-  ></script>
+  <!--
+    For CAPTCHA-enabled journeys: use enterprise.js, not api.js —
+    new Google keys require the Enterprise namespace.
+  -->
+  <script src="https://www.google.com/recaptcha/enterprise.js" async defer></script>
```
ryanbas21: Don't we need the siteKey in here? or is this changed now?

core/journey/stages/_utilities/recaptcha.utilities.ts
```diff
@@ -10,40 +10,91 @@
...
+
+/**
+ * Resolves the active reCAPTCHA namespace. Prefers `grecaptcha.enterprise`
+ * (loaded by enterprise.js) and falls back to the classic `grecaptcha` global
+ * (loaded by api.js or auto-migrated keys). This makes all call sites
+ * transparent to which script the consumer loaded.
+ */
+export function resolveGrecaptcha(): ReCaptchaV2.ReCaptcha {
+  const grecaptcha = window.grecaptcha as ReCaptchaV2.ReCaptcha & {
+    enterprise?: ReCaptchaV2.ReCaptcha;
```
ryanbas21: is `ReCaptchaV2` just a global type we have (I think it is but I have forgotten).

core/journey/_utilities/callback-mapper.svelte
```diff
@@ -58,6 +59,7 @@
     PingOneProtectInitializeCallback,
     PollingWaitCallback,
     ReCaptchaCallback,
+    ReCaptchaEnterpriseCallback,
```
ryanbas21, line 62: So we're now creating a distinct `ReCaptchaEnterpriseCallback`.

Because now in the new "Recaptcha world" there is no significant difference between this for google, this feels like we're diverting our structure away from the actual core Domain (ReCaptcha)

I think maybe what we need to do is align more against HCaptcha vs Google ReCaptcha.

Since GoogleReCaptcha has Enterprise, and others, but they are all under the domain of Enterprise Recaptcha, it can be a bit more resilient to future changes from Google. If they rename, add something, we have just Google based ReCaptcha component that handles that logic, we operate there.

HCaptcha being distinct creates the same separation.

My Concern is that we are creating the separation "Enterprise" vs not, and then if HCaptcha has or adds an Enterprise, are we going to put all Enterprise based logic in the `recaptcha-enterprise` component?

It feels like the abstraction may be on the wrong thing and should be on the Google vs HCaptcha.


This may help clean up some of the mess I originally created when writing this because now the logic for each is coupled (correctly) to the component that it is.

The issue this does create is in this file, since we check the type, we'd need one more layer of logic which says "Which type" of Recaptcha is this.


I think in this case, this second "check" is worth doing for the separation we can create.

Thoughts?

apps/login-app/src/routes/(app)/+page.svelte
```diff
@@ -28,12 +28,15 @@
   const formPostEntryParam = $page.url.searchParams.get('form_post_entry');
   const journeyParam = $page.url.searchParams.get('journey');
   const suspendedIdParam = $page.url.searchParams.get('suspendedId');
+  const captchaModeParam = $page.url.searchParams.get('captchaMode') as
+    | 'normal'
```
ryanbas21: I think maybe we should use the actual names of recaptcha here, `normal` is understandable but I think it may be more clear to use `invisible` and `Not a Robot` then we have a third type which is `Score based` 

If we use `normal` and the world changes to `Score based` we are out of sync with the standard and we break the api.

ryanbas21 (changes requested):
First pass, this will take me a few reads because its ReCaptcha.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
cerebrl: I see you added the `recaptchaAction` to the `callbackMetadata`, but I don't see the use of it in this component. My desire is to use the value that comes from the metadata, rather than relying on this store.

core/journey/journey.store.ts
```diff
@@ -617,3 +626,7 @@ export function initialize(config?: JourneyClientConfig | undefined): JourneySto
 }
 
 export let stack: StackStore;
+
+export const recaptchaActionStore = derived(journeyStore, ($store) => ({
+  recaptchaAction: $store.recaptchaAction ?? '',
+}));
```
cerebrl: Is this still needed now that we have the value woven into the `callbackMetadata`?

cerebrl (changes requested):
I have a few comments around the continued use of the captcha store and if it's still needed. The other captcha comments I'll leave to Ryan as I'm not familiar enough with the spec to weigh in.

apps/login-app/src/routes/(app)/+page.svelte
SteinGabriel: Good point. We could rename `normal` to `checkbox` or `visible` if that's clearer. No breaking behavior change, just a string value in the public interface.

core/journey/_utilities/callback-mapper.svelte
SteinGabriel: Fair point. The component split follows AM's callback structure, not our design. AM sends `ReCaptchaEnterpriseCallback` (different from `ReCaptchaCallback`) for the Enterprise captcha node, and it has a different API too. One component for both would need duck-typing or internal branching, which can be harder to read and test. `hCaptcha` only uses `ReCaptchaCallback`, no Enterprise variant, so it stays in the classic component. If AM adds an hCaptcha Enterprise callback later, it could get its own component or reuse the enterprise one.

If AM is our source of truth for callback components, which it mostly is (one exception aside), a dedicated `ReCaptchaEnterpriseCallback` makes more sense. That said, I get your point about the abstraction belonging at the ReCaptcha vs. hCaptcha level. It really comes down to following AM's structure or creating our own for captcha callbacks.

core/journey/stages/_utilities/recaptcha.utilities.ts
SteinGabriel: Yes,`ReCaptchaV2` comes from the `@types/grecaptcha` npm package. It's a devDependency in `apps/login-app/package.json`.

apps/login-app/src/routes/(app)/+layout.svelte
SteinGabriel: No, for Enterprise the key is passed to `render(element, { sitekey })` or `execute(siteKey)` at render time, the script just needs to be loaded. The `siteKey` value comes directly from `callback?.getSiteKey()`.

core/journey/journey.store.ts
SteinGabriel: Updated. Passed `recaptchaAction` through `callbackMetadata` using the same `initOptions` pattern as `captcha.mode`. `buildCallbackMetadata` now merges `recaptchaAction` from `initializationOptions` into `initOptions` for both captcha callback types. Both components now read `callbackMetadata?.initOptions?.recaptchaAction` reactively. No store subscription and recaptchaActionStore` removed. Thanks!

apps/login-app/src/routes/(app)/+layout.svelte
SteinGabriel: Update: siteKey isn't needed in the API call for reCaptcha v2 only. Enterprise and v3 does need it though. Fixing now.

core/journey/stages/_utilities/recaptcha.utilities.ts
SteinGabriel: Since I was already doing some refactoring, I went ahead and replaced the global window assignments with closures passed directly to `renderCaptcha` / `renderCaptchaInvisible` / `renderEnterpriseCaptcha`. Both `reCaptcha` and `hCaptcha` support a programmatic render() API that accepts function refs, so no window lookup needed, no shared-global risk, no `core/window.d.ts` type.

apps/login-app/src/routes/(app)/+layout.svelte
SteinGabriel: This should be fixed.

core/journey/_utilities/callback-mapper.svelte
ryanbas21: I hear this, but i'm not sure that following AM nodes here is the right decision since the real implementation is more specific to the provider. I'm not hard set on this but just my thoughts

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.utilities.ts
```diff
@@ -0,0 +1,117 @@
...
+
+interface HCaptcha {
+  render: (id: string, options: Record<string, unknown>) => void;
+}
+
+// hcaptcha has no @types package — declare minimal interface and cast via unknown
+const hcaptcha = () => (window as Window & { hcaptcha?: HCaptcha }).hcaptcha;
+
+export function detectEnterpriseProvider(elementClass: string): 'grecaptcha' | 'hcaptcha' {
+  return elementClass === 'g-recaptcha' ? 'grecaptcha' : 'hcaptcha';
```
ryanbas21: can't we change this class at the node level?

core/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -48,31 +46,92 @@
   const recaptchaClass: string =
     callback?.getOutputByName<string>('captchaDivClass', 'h-captcha') ?? 'h-captcha';
 
-  onMount(() => {
+  const captchaProvider: 'hcaptcha' | 'grecaptcha' =
+    recaptchaClass === 'g-recaptcha' ? 'grecaptcha' : 'hcaptcha';
```
ryanbas21, line 52: again i'm concerned that we're relying on this but it's a node level setting if I recall

core/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -48,31 +46,92 @@
...
+
+  // v3 is always Google; v2 provider is derived from captchaDivClass
+  const scriptSrc = isV3
```
ryanbas21: np: maybe an object that maps the urls to the type?

packages/login-widget/src/lib/interfaces.ts
```diff
@@ -55,6 +54,7 @@ export interface Protect {
 }
 
 export interface WidgetConfigOptions {
+  captcha?: { mode?: 'normal' | 'invisible' };
```
ryanbas21: did we want to address this from previous comment?

ryanbas21 (commented):
Just a few comments otherwise looks good

apps/login-app/src/routes/(app)/+page.svelte
SteinGabriel: Renamed `'normal'` to `'visible'`. Chose `'visible'` over `'checkbox'` as it's provider-agnostic and describes the behavior clearly for both Google and hCaptcha. Let me know what you think.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.utilities.ts
SteinGabriel: This function was detecting the captcha provider based on the class name coming from the captcha callback, but `ReCaptchaEnterpriseCallback` is Google-only so the hCaptcha branch in the enterprise component was dead code. 
Removed `detectEnterpriseProvider`, the `HCaptcha` interface, and the hcaptcha render branch. Enterprise always uses `grecaptcha`.

core/journey/callbacks/recaptcha/recaptcha.svelte
SteinGabriel: `captchaDivClass` is read directly from the callback output (`getOutputByName('captchaDivClass')`), so the AM node does control it. We need it not just for the css class but to route between hCaptcha and Google. it drives script loading, and the `hcaptcha.render()` vs `grecaptcha.render()` branching. It's the only signal the callback exposes that we can use to determine the provider.

core/journey/callbacks/recaptcha/recaptcha.svelte
SteinGabriel: Yeah, good call! Replaced the ternary chain with a `CAPTCHA_SCRIPT_URLS` map key. Thanks!

core/journey/callbacks/_effects/captcha.effects.ts
cerebrl, line 1: Since this is a new file, I'd like to move towards the pattern of `utilities` being pure, stateless functions. Since these manage "effects", like reading from the `window` object or interacting with the DOM, let's just reorganize/rename this as an `effect`.

core/journey/callbacks/_utilities/recaptcha-enterprise.utilities.ts
cerebrl, line 1: Do any of these utilities have side-effects? I'm asking because the naming tells me that they are completely effectful.

core/journey/stages/_utilities/recaptcha.utilities.ts
cerebrl, line 1: Same here. Let's move these to `effects` like WebAuthn in the `/stages` directory.

cerebrl (changes requested):
I like the state of this PR. The only changes are the "effectful" utilities. Let's just reorganize/rename these files to reflect the fact that they are not pure, stateless functions.

packages/login-widget/README.md
```diff
@@ -499,6 +499,28 @@ config.set({
...
 ```
 
+### CAPTCHA Configuration
+
+AM does not signal invisible mode in the callback payload for either `ReCaptchaCallback` or `ReCaptchaEnterpriseCallback`. Use the `captcha` option to configure invisible rendering:
+
+```js
+config.set({
+  captcha: {
+    mode: 'invisible', // 'normal' (default) | 'invisible'
```
vatsalparikh: We should update README here if we are changing the name from `normal` to something else as you mentioned in the comment: https://github.com/ForgeRock/forgerock-web-login-framework/pull/478#discussion_r3244297661

core/journey/callbacks/recaptcha/recaptcha.svelte
```diff
@@ -48,31 +46,92 @@
...
       return;
     }
-    if (callback) {
-      window.frHandleCaptchaError = handleCaptchaError(callback);
-      window.frHandleCaptcha = handleCaptchaToken(callback);
-      window.frHandleExpiredCallback = function handleExpiredCallback() {
+
+    const mountMode =
+      ((callbackMetadata?.initOptions?.mode as string | undefined) ?? 'visible') === 'invisible'
+        ? 'invisible'
```
vatsalparikh: I think we should define a type for captcha mode, maybe `CaptchaMode`. It seems like this value is referenced across lots of files. Defining it as a type will avoid typos and `as` casting, and provide a single reference across multiple files.

core/journey/callbacks/_effects/captcha.effects.ts
SteinGabriel: Updated. Split all three files following the WebAuthn pattern. Pure functions remain in `*.utilities.ts`; DOM/window effects moved to `*.effects.ts`.  Thanks!

vatsalparikh (commented):
**Provided code suggestions in respective lines. Below are some PR-wide suggestions**

- File structure of utilities and effects files is confusing. We should consolidate this logic under callbacks/ folder because this is callback related logic
	- We have effects and utilities under recaptcha-enterprise. We don't have this pattern, those should be moved out to dedicated _effects and _utilities folders.
	- Then we have effects and utilities under stages for recaptcha
	- And we have effects and utilities under journey/_utilities. 
	- Any effects files should be under its own _effects folder and _utilities under its own _utilities folder.
- For 1 pending check, please approve or deny the changes in chromatic before merging. I've found UI regressions from my PRs here before.
Image

apps/login-app/src/routes/(app)/+page.svelte
ryanbas21: Why not just use the names of the recaptcha instead? "Not a robot" | "invisible" | "score" ?

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.effects.ts
```diff
@@ -0,0 +1,81 @@
...
+    try {
+      const token = await grc.execute(siteKey, { action });
+      callback?.setResult(token);
+      try {
+        callback?.setAction(action);
+      } catch {
+        // AM node may not expose the action input — non-fatal, token already set
+      }
+    } catch (err) {
+      console.error('[recaptcha-enterprise] execute() threw:', err);
```
ryanbas21: do we want to log out consoles in production code?

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
```diff
@@ -0,0 +1,116 @@
...
+        : 'visible';
+
+    try {
+      await loadEnterpriseScript({
+        apiUrl,
+        siteKey,
+        mode: mountMode,
+      });
+    } catch (err) {
+      console.error('[recaptcha-enterprise] script load failed:', err);
```
ryanbas21: console in prod code again

apps/login-app/src/routes/(app)/+page.svelte
SteinGabriel: Because the captcha mode option controls both google reCaptcha and hCaptcha. "Not a robot" is google's checkbox UI copy, not a mode name, and hCaptcha has no equivalent term. 'score' isn't a valid value here either, v3/score-based mode is detected directly from the AM payload (reCaptchaV3: true output) and never needs to be set by the consumer. So the only real choice a consumer makes is how the widget renders: does it show a visible checkbox or execute silently in the background? 'visible' | 'invisible' describes that behavior precisely and stays correct and consistent regardless of the provider.j

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.effects.ts
SteinGabriel: No, that was a debug `console.error`. Removed, thanks.

core/journey/callbacks/recaptcha-enterprise/recaptcha-enterprise.svelte
SteinGabriel: Removed. Thank you.

core/journey/callbacks/recaptcha/recaptcha.svelte
SteinGabriel: Good call. I added `CaptchaMode = 'visible' | 'invisible'` type in `journey.interfaces.ts`. Both callback components now import and use it. This also fixed a TS error where `initOptions` caused the reactive `captchaMode` to widen to `string`, which broke the `loadEnterpriseScript` `mode` param. Thanks!

packages/login-widget/README.md
SteinGabriel: That's updated. README now reflects `'visible'` | `'invisible`. Thanks for pointing this out.

core/journey/_utilities/callback-mapper.svelte
SteinGabriel: I see your point. The existing `reCaptchaCallback` naming can be specially misleading/confusing since it's used by both reCaptcha and hCaptcha, even though that's the callback name returned by the Captcha node. `reCaptchaEnterprise` is more consistent with the journey node and its intended use.

I'll spend some time splitting the callback components into `hcaptcha.svelte` and `recaptcha.svelte` (classic and enterprise) to see if organizing them by provider makes more sense. Thanks!

cerebrl (approved):

apps/login-app/src/routes/e2e/widget/inline/+page.svelte
```diff
@@ -24,6 +24,9 @@
   let authIndexValueParam = $page.url.searchParams.get('authIndexValue');
   let journeyParam = $page.url.searchParams.get('journey');
   let recaptchaParam = $page.url.searchParams.get('recaptchaAction');
+  const captchaModeRaw = $page.url.searchParams.get('captchaMode');
+  let captchaModeParam: 'visible' | 'invisible' | null =
```
vatsalparikh: I think `let captchaModeParam: 'visible' | 'invisible' | null =` is redundant. We can just do

```
  const captchaModeParam =
    captchaModeRaw === 'visible' || captchaModeRaw === 'invisible' ? captchaModeRaw : null;
```

And be consistent in other files too like
- `apps/login-app/src/routes/(app)/+page.svelte`
- `apps/login-app/src/routes/e2e/widget/modal/+page.svelte`

core/journey/stages/_effects/captcha.effects.ts
```diff
@@ -0,0 +1,14 @@
...
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ **/
+
+export {
+  renderCaptcha,
+  renderCaptchaInvisible,
+  resolveGrecaptcha,
+} from '$journey/_utilities/captcha.effects';
```
vatsalparikh: This file should not exist. It is re-exporting things from journey utilities and doing nothing else. We should avoid this misdirection

vatsalparikh (approved):
**Current review non-blocking comments**
- We should document the recaptcha config gotchas because they deviate from official recaptcha enterprise docs. This should be documented in this repo's README file so the knowledge doesn't stay tribal.
- Some new line-specific non-blocking comments


**Previous review non-blocking comments**
- File structure of utilities and effects files is confusing. We should consolidate this logic under callbacks/ folder because this is callback related logic
We have effects and utilities under recaptcha-enterprise. We don't have this pattern, those should be moved out to dedicated _effects and _utilities folders.
Then we have effects and utilities under stages for recaptcha
And we have effects and utilities under journey/_utilities.
Any effects files should be under its own _effects folder and _utilities under its own _utilities folder.
- For 1 pending check, please approve or deny the changes in chromatic before merging. I've found UI regressions from my PRs here before.

apps/login-app/src/routes/e2e/widget/inline/+page.svelte
SteinGabriel: Yeah, that's true. Updated it in all 3 files. Thanks!

core/journey/stages/_effects/captcha.effects.ts
SteinGabriel: Good catch. That looks like a leftover from recent refactoring. Deleted `stages/_effects/captcha.effects.ts`. Merged its tests into `_utilities/captcha.effects.test.ts` where the actual functions live.

ryanbas21 (approved):

PR #489, opened 2026-05-04
fix(mfa-step): create a new step for handling MFA set up page text output callback script

vatsalparikh:
## JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-4918

## Description
This PR addresses scripted nodes for MFA during 2FA registration process. More details in the ticket

## Reasoning
- MFA enrollment, Get authenticator, and App links screens are UI concerns only, there is no data transformation or extraction happening here. That's why catching these scripts and implementing them in login framework and not in Ping SDK.
- MFA enrollment, Get authenticator, and App links screens are more than callbacks, they are a combination of multiple callbacks like text output, hidden value, confirmation callback. That's why I implemented them as a separate stage.
- I was itching to use custom stages and callbacks here, but because the feature is experimental I did not take that route.
- For rendering the screens, I've tried to reuse Alert, Text and other common components, so the styling is slightly off but follows the styling in the repo for Alert and Text.
- Open to suggestions on how to better detect all three screens, and whether we should separate out various steps in the mfa enrollment or if you have a better idea on how to handle scripted nodes safely (to avoid xss).

## How to test
- Make sure the FR_REALM_PATH=/ and FR_AM_WELLKNOWN_URL=https://openam-sdks.forgeblocks.com/am/oauth2/realms/root/realms/root/.well-known/openid-configuration are configured so that you are in the root realm and not in the alpha realm
- Start up the app and login with your admin credentials on https://localhost:8443. These are the credentials you use to login to [openam-sdks tenant](https://openam-sdks.forgeblocks.com/). Do not use a user you've created in the openam-sdks tenant because those users are not admins.
- Once you login you should see the Set up screen. Click on 'Set up', then on 'Next', then on 'Download the app'. All three screens were showing text and now show proper nodes.
- You can refer to the recording, which shows both text scripts before the changes and the nodes after the changes.

## Recording
- Before changes
https://github.com/user-attachments/assets/e8bdfecd-2eba-472d-80fe-8ab272ffa85b

- After changes
https://github.com/user-attachments/assets/e13e6eb4-91e9-451f-ad89-d297ee002dc5

## +error.svelte file
I forgot to push the +error.svelte file from Gabriel's comment in journey client PR. Since it's a generic error file I have attached it in this PR.
https://github.com/ForgeRock/forgerock-web-login-framework/pull/468#discussion_r3190857362


core/journey/_utilities/map-stage.utilities.ts
```diff
@@ -33,9 +33,11 @@ type StageTypes =
   | typeof Registration
   | typeof Login
   | typeof Generic
+  | typeof MfaEnrollment
```
SteinGabriel, line 41: It appears that `MfaEnrollment` is not imported in this file.

core/journey/_utilities/map-stage.utilities.ts
```diff
@@ -95,5 +97,24 @@ export function mapStepToStage(currentStep: StepTypes): StageTypes | Component {
     return EmailSuspend;
   }
 
+  const hiddenValueCallbacks = currentStep.getCallbacksOfType(
+    callbackType.HiddenValueCallback,
+  ) as HiddenValueCallback[];
```
SteinGabriel, line 107: Same for `HiddenValueCallback` and `TextOutputCallback`.

core/journey/_utilities/map-stage.utilities.ts
```diff
@@ -95,5 +97,24 @@ export function mapStepToStage(currentStep: StepTypes): StageTypes | Component {
...
+
+  // App links screen: type-4 script targets the callback's XUI-rendered DOM element directly
+  const textOutputCbs = currentStep.getCallbacksOfType(
+    callbackType.TextOutputCallback,
+  ) as TextOutputCallback[];
+  const hasAppLinksScript = textOutputCbs.some(
+    (cb) =>
+      cb.getMessageType() === '4' && cb.getMessage().includes('document.getElementById("callback_'),
+  );
+  if (hasAppLinksScript) return MfaEnrollment;
```
SteinGabriel, line 123: It'd probably be worth adding tests for `hasMfaHidden` and `hasAppLinkScript`.

SteinGabriel (changes requested):
Initial scan.

core/journey/_utilities/map-stage.utilities.ts
vatsalparikh: Sorry, looks like rebase didn't go through as expected. I was rebasing with all changes made to journey client from main. It's surprising that the linter didn't complain.

Thanks, it is fixed now.

core/journey/_utilities/map-stage.utilities.ts
vatsalparikh: Sorry, looks like rebase didn't go through as expected. I was rebasing with all changes made to journey client from main. It's surprising that the linter didn't complain.

Thanks, it is fixed now.

core/journey/_utilities/map-stage.utilities.ts
vatsalparikh: They are already tested as part of stories. I added tests anyway, thanks!

core/locales/us/en/index.json
```diff
@@ -11,7 +11,12 @@
   "confirmPassword": "Confirm password",
   "constraintViolationForPassword": "Password does not meet the requirements.",
   "constraintViolationForValue": "Value does not meet the requirements.",
+  "continueButton": "Continue",
   "continueWith": "Continue with ",
+  "downloadTheApp": "Download the app",
+  "getAuthenticatorApp": "Set up the ForgeRock Authenticator",
+  "getAuthenticatorAppDescription": "To get started, you need to register your device using the ForgeRock Authenticator app.",
+  "getAuthenticatorAppLinks": "Get the app from the <a href='https://itunes.apple.com/app/forgerock-authenticator/id1038442926' target='_blank'>Apple App Store</a> or on <a href='https://play.google.com/store/apps/details?id=com.forgerock.authenticator' target='_blank'>Google Play Store</a>",
```
SteinGabriel: We should probably add `rel='noopener norefereer'` to the anchor tags being injected into the DOM, to avoid the risk of tab-nabbing and URL leaking. 

Update: I checked if there was any sanitization around this, and apparently `rel` would be removed because only `target`, `href`, and `title` are in the xss allowlist for the `a` tag by default, as described in [their docs](https://github.com/leizongmin/js-xss/blob/master/README.md#whitelist).

That said, we'll probably need to pass an allowlist containing `rel` to the sanitize function options, if we'd like to address this in this PR. 
Let me know what you think.

vatsalparikh: Thanks for looking into xss and docs on which attributes are supported by xss, extremely helpful!

While we can support xss whitelist I feel like we can remove the target attribute and avoid introducing rel. It does mean that users will have to right click to open that link in a new tab, but I think it's worth it to avoid adding any whitelist at all, if possible.

And I don't see target being used in the other anchor tags, only in three places where I added code, and adding target attribute doesn't bring much benefit compared to the code we need to write to avoid xss.

Do you think it's okay to remove target attribute?

SteinGabriel: Yeah I get where you're coming from, but I think dropping `target="_blank"` creates a real UX regression here. These links appear during MFA enrolment, the user needs to install the authenticator app and then come back to finish the flow. If clicking the app store link navigates away from the login page, the journey session is likely lost and they'd have to restart authentication. 

I believe adding `rel` to the anchor tag sanitization allowlist should be simple enough to avoid the UX regression. Especially because there is already a default allowlist in place (`target`, `href`, `title`), we'll just be extending it by adding the `rel` property.

core/sdk.config.ts
```diff
@@ -7,9 +7,11 @@
  *
  **/
 
-import { CallbackType, Config, type FRCallback } from '@forgerock/javascript-sdk';
+import { Config } from '@forgerock/javascript-sdk';
```
ryanbas21, line 10: are we still on legacy sdk?

vatsalparikh: We need this to support oauth and user store, both are still using legacy sdk. Once we replace those, we can get rid of the entire sdk.config.ts file

core/journey/_utilities/map-stage.utilities.ts
```diff
@@ -95,5 +102,24 @@ export function mapStepToStage(currentStep: StepTypes): StageTypes | Component {
...
+  });
+  if (hasMfaHidden) return MfaEnrollment;
+
+  // App links screen: type-4 script targets the callback's XUI-rendered DOM element directly
+  const textOutputCbs = currentStep.getCallbacksOfType(
+    callbackType.TextOutputCallback,
+  ) as TextOutputCallback[];
+  const hasAppLinksScript = textOutputCbs.some(
+    (cb) =>
+      cb.getMessageType() === '4' && cb.getMessage().includes('document.getElementById("callback_'),
```
cerebrl: I'm worried that the second condition, the `cb.getMessage().includes ... `, is not specific enough. It feels like that string would be commonly found in other scripts. Is there something more explicitly unique that we can detect to ensure this doesn't result in false positives?

cerebrl (changes requested):
This looks good, other than the condition I commented on.

core/journey/_utilities/map-stage.utilities.ts
vatsalparikh: Agreed. I was going back and forth on this decision on whether to make it specific to app urls. I've made it more specific by looking for app store or play store urls. That's the most specific thing about this step that displays app links and urls. Thank you!

```
      (cb.getMessage().includes('itunes.apple.com') ||
        cb.getMessage().includes('play.google.com')),
```

core/locales/us/en/index.json
vatsalparikh: Agreed, added a whitelist property and rel attribute, thank you 👍

cerebrl (approved):

SteinGabriel (approved):
The changes look good and work as expected.
It seems like a changeset is needed, but other than that it looks ready to merge.

vatsalparikh:
> The changes look good and work as expected. It seems like a changeset is needed, but other than that it looks ready to merge.

Added changeset, thanks!

PR #492, opened 2026-05-07
fix(eslint-import): prevent type-only import lint failures

vatsalparikh:
## Description

Fixes CI lint failures caused by type-only imports in the SDK config. Updates the import style in the SDK config file and improves the pre-commit linting step so these issues are auto-fixed before pushing.

### DIFF
--- a/core/sdk.config.ts
+++ b/core/sdk.config.ts
@@ -7,8 +7,10 @@
  *
  **/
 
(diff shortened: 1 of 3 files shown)


SteinGabriel (approved):
Fix looks good. Tested it locally and it worked as expected.

PR #494, opened 2026-05-13
fix(custom-components): address issues found during QA for custom components

vatsalparikh:
# JIRA ticket
https://pingidentity.atlassian.net/browse/SDKS-4703

## Description
This PR addresses issues found during QA for custom components and ping-lf CLI tool (and journey client migration regression). Each commit addresses an issue / feature / regression. This should help isolate the PR review for each issue / feature / regression.

## Issues addressed

- **Migrate templates:** two templates in ping-lf cli tool were using legacy sdk, migrated them to journey client
- **Installation**: login-framework-cli is specified as workspace instead of an actual version number. Need to either remove the exclusion for tools/ folder from source repo or point to npm version 0.0.0-beta.0, so that the cloned repo can perform update command. This is necessary because consumers who use `ping-lf update` don't have ping-lf installed in workspace, and that happens because we are excluding `tools/` folder in exclusions.ts file in login-framework
- **`--version` flag**: ping-lf init <dir> --version v1.2.0 and ping-lf update --version v1.2.0 ignored the value, printed the CLI's own version and exited. The --version flag was defined in three places: as a global flag on the root ping-lf command and as a subcommand option on both init and update. The global flag won during parsing, so the subcommand value was never read. Renamed the subcommand option to `--tag`, since the value is a Git release tag of the framework repo. Tested with both commands and the .generator-version file was updated with the appropriate login-framework version.
- **Empty commit hash**: I confirmed with Gabriel that empty commit hash is dead code. However, the .generator-version file which helps developers know about the login framework version they are using can be confused with the existence of commitHash variable which is empty, so I removed it entirely from the codebase to avoid confusing users.
- **ping-lf generate stage**: Stage names in AM are arbitrary strings. They can contain spaces (e.g. "My Login Stage"). The CLI was substituting the raw name directly into TypeScript function names in the generated utilities files, producing formatMy Login StageLabel which doesn't compile. The fix derives a PascalCase identifier from the kebab slug ("MyLoginStage") via a new __COMPONENT_NAME_PASCAL__ placeholder, used only in identifier positions. The raw name is preserved everywhere else so AM stage matching continues to work correctly.
- **parseAcceptedProps**: parseAcceptedProps function used a regex and had a bug that caused the callback to appear twice where the regex matched an example snippet inside a JSDoc comment. Fixed by replacing the regex with a Svelte compiler AST.
- **$login-framework alias:**
The README tells developers to import only from $login-framework and never reach into internal aliases. The templates don’t follow this pattern. Updated templates and tools/cli/tsconfig to use $login-framework alias.
This introduced a circular import issue: choice.svelte  → imports $login-framework (login-framework.ts) → exports CallbackMapper from callback-mapper.svelte → imports custom-registry.ts → imports choice.svelte  ← back to start.
To fix this (I’ve implemented the first fix)
we can wrap the callbacks inside a function so they are lazily read during render instead of at runtime
Or we can remove the callback mapper export from login-framework and document it in readme that CallbackMapper is the only export in $login-framework that imports custom-registry.ts. Remove it from the barrel and the chain breaks entirely. Stage authors import it directly from $journey/_utilities/callback-mapper.svelte instead. No cycle, no workaround needed.

## Features added
- **New templates for tests**: Add new templates for both callbacks and stages for story.svelte, [stories.js](http://stories.js/), and [mock.ts](http://mock.ts/) files, so that when you now run the ping-lf command it generates story, stories, mock files along with the custom component’s .svelte file

## Journey Client regression
- While using custom callbacks and custom components in sample app, a journey client regression was found. Calling configuration() without a journeyClient crashed the sample app. That first call to configuration() crashed. With no journeyClient, setJourneyClientConfig(undefined) threw, so the app died before the second set() ever ran.
- The fix is to initialize the journey store even with an undefined journeyClient. When journeyClient is undefined, the consumer app does not crash, only provides an error when the actual journey begins. It also continues to allow setting journeyClient either during configuration() call or configuration().set() call.


packages/login-widget/package.json
```diff
@@ -43,7 +43,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.23.9",
-    "@forgerock/login-framework-cli": "workspace:*",
+    "@forgerock/login-framework-cli": "0.0.0-beta.0",
```
ryanbas21: can you explain this change? we shouldn't lock it to this version, right?

tools/cli/README.md
```diff
@@ -157,6 +157,30 @@ pnpm test
...
+
+By default, `packages/login-widget` pins `@forgerock/login-framework-cli` to a specific npm version. If you want to develop the CLI and test changes against the widget in the same repo, you can switch to the local workspace package instead.
+
+**1. Update `packages/login-widget/package.json`:**
+
+```diff
+-    "@forgerock/login-framework-cli": "0.0.0-beta.0",
++    "@forgerock/login-framework-cli": "workspace:*",
+```
+
```
ryanbas21: this seems to contradict the change above?

ryanbas21 (changes requested):

packages/login-widget/package.json
vatsalparikh: The thing is, when we have workspace:* in the package.json and we clone the folder with ping init command, it creates the new project in a new directory. 

The project build fails for the custom login framework because we are excluding the /tools/cli folder in exclusions.ts file: https://github.com/ForgeRock/forgerock-web-login-framework/blob/eef8b31b713b5f88f06375d76970ad87235bc6ac/tools/cli/src/config/exclusions.ts#L24

We can either let the developer copy over the tools/cli folder so it can be referenced through workspace or we can provide a specific npm version. 

I am pinning this version because that's the one available on npm. 
We can specify `latest`, although `latest` is worse compared to a specific version in the sense that it's unstable. So I specified a particular version, and we have only 0.0.0-beta.0 available right now. 

It just creates a friction for the developer who has copied this repo. By pinning the version, the developer can get started easily without any errors. And now that we have ping-lf already published on npm, it makes sense to reference an npm version instead of workspace.

tools/cli/README.md
vatsalparikh: This is for local development only as explained in the above lines before the diff. If we are in the same repo and actively developing ping-lf we need to use workspace. I can remove this from readme if it is confusing.

```
### Using `workspace:*` instead of a published version

By default, `packages/login-widget` pins `@forgerock/login-framework-cli` to a specific npm version. If you want to develop the CLI and test changes against the widget in the same repo, you can switch to the local workspace package instead.

**1. Update `packages/login-widget/package.json`:**

 
Also update the `prebuild` script in the same file — with `workspace:*` the symlink points to `tools/cli` directly, so the path must reference the compiled output rather than the pre-built `.mjs` wrapper:


**2. Re-run `pnpm install`** from the repo root to update `pnpm-lock.yaml`.

> **Note:** Do not commit these changes. Switch back to the pinned npm version before pushing.
```

packages/login-widget/package.json
ryanbas21: I think if this fails, then the CLI needs to handle this appropriately.

I don't think pinning the version is the right call.

We can either parse or install it directly during the process of the clone.

tools/cli/README.md
vatsalparikh: Had a conversation on Slack. Going back to workspace:* so the above comment is not necessary anymore and has been removed!

packages/login-widget/package.json
vatsalparikh: Had a conversation on Slack. Going back to workspace:* instead of a pinned version to avoid stale installations and version mismatch for ping-lf.

vatsalparikh: The issue with using workspace is now we need to either export the `tools/` folder or somehow specify a version of ping-lf that the cloned repo can use.

I think we can just remove `tools/` folder from exclusions file. This way, the cloned repo will always have ping-lf and it will work with workspace as expected. Otherwise I don't see how workspace:* will help. While the source (login-framework) repo will use workspace:*, the cloned repo will still need some version either from npm or locally. If it reads from npm, there is already divergence. If we allow tools/ folder, that's the only way those versions will match.
I do feel the tug of war where ping-lf now becomes a requirement, not optional

SteinGabriel: @vatsalparikh Is this the error you saw when building the generated project? https://github.com/ForgeRock/forgerock-web-login-framework/pull/500#discussion_r3291442757

I believe you're right about not using "workspace". Why can't we use `"@forgerock/login-framework-cli": "^0.0.0-beta.0"` instead? That way it allows updates to minor/patch versions. It may also be possible to update this via the `update` CLI command. 
Also, I think we need to change the login-widget prebuild script to `"prebuild": "node node_modules/@forgerock/login-framework-cli/dist/scripts/generate-registry.js ../.."`  to use the `generate-registry` script from the installed package, instead of workspace, which does not exist in the generated project.

vatsalparikh: Yeah, if we exclude tools/cli folder, then the generated project relies on workspace but doesn't find it. That's why I removed tools/ folder from exclusions. I agree, I initially implemented the fix with "^0.0.0-beta.0". To me, it sounds the most stable and robust fix. Any developer who is using it knows that they are using a ping-lf version from npm repo. If they wish to make changes to ping-lf locally and use that, then they can add to workspace themselves, so it is self-explanatory. Would be great to discuss this on and decide on the path: whether we want to go with "^0.0.0-beta.0" or "tools/ folder export" as a solution.

SteinGabriel: The only thing for me about removing `tools/` from the exclusions is that the tools folder shouldn't be necessary to build and run custom components. Initially I had the `generate-registry.js` script within the `/core` folder instead of the `/cli` folder, and I wonder if that was the right call in the end. Anyway, I agree we should discuss this, maybe on Tuesday when Ryan is back.

packages/login-widget/src/lib/_utilities/api.utilities.ts
```diff
@@ -137,7 +139,9 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
          * Initialize the stores and ensure both variables point to the same reference.
          * Variables with _ are the reactive version of the original variable from above.
          */
-        journeyStore = initializeJourney(setOptions?.journeyClient);
+        if (setOptions?.journeyClient) {
+          journeyStore = initializeJourney(setOptions.journeyClient);
+        }
```
SteinGabriel: Now that `journeyStore` is initialized only when `journeyClient` is passed into the store,  `journeyStore` is returned as `undefined` when `journeyClient` is not passed in. I think this can create a few issues when `configuration()` is called without `journeyClient`. I also noticed that we didn't have tests for `api.utilities.ts`. I added a few tests for these issues, but I wasn't able to create a commit since these tests fail. Let me know if you'd like me to push them using `--no-verify` tag.

tools/cli/src/commands/generate.ts
```diff
@@ -34,6 +34,17 @@ function toSlug(name: string): string {
...
+/**
+ * Converts a kebab-case slug to PascalCase, safe for use as a TypeScript identifier.
+ * Examples: "my-login-stage" → "MyLoginStage", "otp-login" → "OtpLogin"
+ */
+function toPascalCase(slug: string): string {
+  return slug
+    .split('-')
+    .map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+    .join('');
+}
```
SteinGabriel: There's an existing `toPascalCase()` function in `registry.ts`. Can we move this function to a shared utility file to avoid duplication?

 Also, since it receives a slug here, I think it could be an issue when the slug name is a single word. 
For example: `toSlug(DefaultLogin)` → defaultlogin → `toPascalCase(defaultlogin)` → Defaultlogin, where it should be DefaultLogin (capital 'D' & 'L') instead.

tools/cli/templates/callback/__COMPONENT_SLUG__.svelte
```diff
@@ -46,7 +45,7 @@ Custom callback component. Replace this description with your own.
    * The widget's resolved style configuration (colors, logos, labels, etc.).
    * Use this to keep your custom callback visually consistent with the theme.
    */
-  export const style: z.infer<typeof styleSchema> = {};
```
SteinGabriel, line 49: Since developers don't need to declare these props when they are not used within the components, I think we are teaching the wrong patterns by keeping them in the template. I believe we should have them commented out with a disclaimer instead. 

Something like, but keeping each prop's comment block:
```js
  export let callback: BaseCallback;
  // Optionally declare any of these if your component needs them:
  // export let selfSubmitFunction: Maybe<SelfSubmitFunction> = null;
  // export let stepMetadata: Maybe<StepMetadata> = null;
  // export let callbackMetadata: Maybe<CallbackMetadata> = null;
  // export let style: StyleObject = {};
```

SteinGabriel (changes requested):

packages/login-widget/src/lib/_utilities/api.utilities.ts
vatsalparikh: Thank you, that was interesting!

I've made changes to `setJourneyClientConfig` function so that setting journey client returns undefined but journey store always initializes. This avoids the app from crashing. I am still returning undefined from `setJourneyClientConfig`. So this works with existing setup, the only thing is the sample app needs a wellknown config which is anyways necessary for login widget now.

I tested it with sample app all three scenarios. All three cases work well. The sample app does not crash anymore just because journey client is undefined.
- no journey client
- journey client only defined in configuration().set() part of login widget config
- journey client only defined in in configuration() part of login widget config

Thanks for trying to add tests, I've added tests, but go ahead and commit them and I'll check if they are still needed 👍 

I've renamed api.utilities.ts file to widget.api.ts file because this file is not utilities, it is public API. We are mutating stuff, making http calls, and such.

tools/cli/src/commands/generate.ts
vatsalparikh: Thanks, makes sense! 

I moved the common function to utils and am passing a `name` to the `toPascalCase()` function.

tools/cli/templates/callback/__COMPONENT_SLUG__.svelte
vatsalparikh: Nice idea! Done!

packages/login-widget/package.json
SteinGabriel: Tested it locally and was able to install dependencies and run login-app on the generated project with no errors. I still think the `tools/` folder shouldn't be part of the generated project, which can be discussed further, but this fixes the issue. Thank you!

cerebrl (commented):
This all looks good to me. I'll leave the official approval to Gabriel and Ryan as they are more familiar with this particular section of the code.

packages/login-widget/package.json
vatsalparikh: Had a long conversation in google meet with Ryan and Gabriel. 

This is what we landed on: We avoid any dependencies on CLI tool in the login framework, so remove prebuild and workspace dependencies. Ryan suggested we use vite plugin for any tool dependencies.

Brainstormed with Gabriel. Gabriel suggested that we completely stop building registry from the ping-lf CLI tool. This way, CLI tool only `generates` the svelte files and that's the extent of it. That's the only responsibility of the CLI tool. Registering the components is handled by login-widget or login-app through vite plugin.

ryanbas21 (approved):
Looks good, great work.

SteinGabriel (approved):
Changes look good! Tested new project init, dependency install, and custom component generation. All built fine and worked as expected. Thanks @vatsalparikh! 

One minor note (not a blocker): `toPascalCase` is now duplicated in tools/cli/src/utils.ts` and `core/journey/_utilities/registry/registry.ts`. Duplication makes sense to avoid a core/tools dependency, so maybe we should just add a comment on both explaining it's intentional and the reason why.

vatsalparikh:
> Changes look good! Tested new project init, dependency install, and custom component generation. All built fine and worked as expected. Thanks @vatsalparikh!
> 
> One minor note (not a blocker): `toPascalCase` is now duplicated in tools/cli/src/utils.ts`and`core/journey/_utilities/registry/registry.ts`. Duplication makes sense to avoid a core/tools dependency, so maybe we should just add a comment on both explaining it's intentional and the reason why.

Thanks! I have added a comment to both functions mentioning the reason for duplication!

PR #500, opened 2026-05-18
feat(cli): add mcp subcommand to expose CLI commands as MCP tools

ryanbas21:
## Summary

- Adds a \`mcp\` subcommand to \`@forgerock/login-framework-cli\` — when invoked, the binary boots a local MCP server over stdio
- Exposes five typed tools (\`init\`, \`generate_callback\`, \`generate_stage\`, \`update\`, \`list_releases\`) backed by the same Effect functions the CLI already uses — no business logic duplication
- All tool errors produce structured, human-readable messages via per-tag pattern matching on \`Data.TaggedError\` payloads
- \`Logger.prettyLogger({ stderr: true })\` routes all Effect logs to stderr, keeping stdout clean for the MCP JSON-RPC stream
- Adds \`@forgerock/login-framework-cli\` as a local workspace link in root \`dependencies\` and \`pnpm.overrides\`

## Architecture

```
src/
├── main.ts     # ternary: process.argv[2] === 'mcp' → Layer.launch(ServerLayer)
│               # else → CLI pipeline; single NodeRuntime.runMain for both paths
├── mcp.ts      # McpServer + Toolkit definitions + tool handlers + exported ServerLayer
└── commands/
    ├── init.ts      # exports initProject()
    └── generate.ts  # exports scaffoldComponent()
```

The \`mcp\` subcommand stub is registered with \`@effect/cli\` only so it appears in \`--help\` output. The actual dispatch happens via \`Layer.launch(ServerLayer)\` at the top level of \`main.ts\`, outside the CLI's fiber scope, which avoids nesting a second \`Layer.launch\` inside the already-running Effect runtime.

## Usage

```jsonc
// Claude Desktop / .claude/mcp.json
{
  "mcpServers": {
    "ping-lf": {
      "command": "npx",
      "args": ["@forgerock/login-framework-cli", "mcp"],
      "cwd": "/path/to/your/login-project"
    }
  }
}
```

## Test Plan

- [ ] `pnpm --filter @forgerock/login-framework-cli exec vitest run` — all tests pass
- [ ] TypeScript: `tsc --noEmit` — zero errors
- [ ] Manual: `node dist/src/main.js mcp` then send `{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}` — returns all 5 tools with correct annotations
- [ ] Manual: send `initialize` request — responds with capabilities object including `tools` capability
- [ ] Manual: `node dist/src/main.js --help` — `mcp` appears in COMMANDS list


cerebrl (commented):
This looks cool from a cursory scan. I will revisit this soon to for a more in-depth review. I'd like to see @vatsalparikh and @SteinGabriel look at this as they may have more feedback.

tools/cli/src/main.ts
```diff
@@ -1,93 +1,99 @@
...
 import { generateCommand } from './commands/generate.js';
 import { initCommand } from './commands/init.js';
 import { releasesCommand } from './commands/releases.js';
 import { updateCommand } from './commands/update.js';
 import { GithubReleaseLayer } from './services/release.js';
 
-// Read the version from package.json at runtime so it stays in sync
-// with the published package version after changesets bumps it.
-const { version } = createRequire(import.meta.url)('../package.json') as { version: string };
+if (process.argv[2] === '--mcp') {
```
vatsalparikh: We should follow the existing format for all our commands, like `init`, `update`, `releases`.

Something like

```
export const mcpCommand = Command.make('mcp', {}, () =>
  Effect.sync(() => runMcpServer())
).pipe(
  Command.withDescription('Start as an MCP server over stdio.')
);
```

vatsalparikh (commented):
Very early initial test:

Tried creating mcp server. Because of hard coded cwd, faced issues using the server.

Used a directory param, and things are working well.

tools/cli/src/main.ts
ryanbas21: thanks, fixed and the above dir issue also and pushed

package.json
```diff
@@ -99,7 +99,11 @@
       "picomatch@2": "2.3.2",
       "picomatch@4": "4.0.4",
       "uuid": "^14.0.0",
-      "yaml": ">=2.8.3"
+      "yaml": ">=2.8.3",
+      "@forgerock/login-framework-cli": "link:tools/cli"
     }
+  },
+  "dependencies": {
+    "@forgerock/login-framework-cli": "link:tools/cli"
```
SteinGabriel: Got this error running `pnpm install`  on the generated project: 
```
 ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In packages/login-widget: "@forgerock/login-framework-cli@workspace:*" is in the dependencies but no package named "@forgerock/login-framework-cli" is present in the workspace
```


Not sure if it's pre-existing or caused by adding `login-framework-cli` as a dependency.

tools/cli/src/services/registry.ts
```diff
@@ -0,0 +1,296 @@
...
+ * spawning. Validation errors across multiple components are collected and
+ * reported together.
+ */
+export const runRegistryScript = (projectDir: string) =>
+  Effect.gen(function* () {
+    const fs = yield* FileSystem.FileSystem;
+    const path = yield* Path.Path;
+    const stageDir = path.join(projectDir, 'experimental', 'custom', 'stages');
+    const callbackDir = path.join(projectDir, 'experimental', 'custom', 'callbacks');
+    const registryDir = path.join(projectDir, 'core', 'journey', '_utilities');
```
SteinGabriel: `custom-registry.ts` now lives in `core/journey/_utilities/registry/`, so `path.join()` needs to be updated to include a 'registry' segment.

tools/cli/src/services/registry.ts
```diff
@@ -0,0 +1,296 @@
+import { FileSystem, Path } from '@effect/platform';
+import { Console, Effect } from 'effect';
+
+import { RegistryScanError } from '../errors.js';
```
SteinGabriel, line 4: Looks like `RegistryScanError` is being exported from `core/journey/_utilities/registry/registry.ts` instead of `../errors.js`. I believe It should be exported from `errors.js` for consistency, but either way this needs to be updated.

tools/cli/src/commands/init.ts
```diff
@@ -32,6 +35,91 @@ To scaffold your first custom component:
...
+    // Run the canonical registry script so the file is immediately present
+    // with the correct CustomRegistryEntry format. The empty callbacks/ and
+    // stages/ dirs produce empty registries — no custom components yet.
+    yield* Console.log('Generating custom component registry...');
+    yield* runRegistryScript(resolvedDir);
+
+    // ── 6. Write .generator-version ────────────────────────────────────────
+    yield* writeVersion(resolvedDir, {
+      version: resolvedVersion,
+      commitHash: '',
```
SteinGabriel: `commitHash` was removed from `GeneratorVersionSchema` at some point, so `commitHash` should not be included here. The same applies in `commands/update.ts` and `mcp.ts`.

tools/cli/src/services/registry.ts
SteinGabriel: By the way,  this looks like the file moved to `core/journey/_utilities/registry/registry.ts` in the QA ticket. If that's the case, this file should be deleted.

SteinGabriel (changes requested):

tools/cli/src/commands/generate.ts
```diff
@@ -112,20 +113,15 @@ function scaffoldComponent(type: 'callback' | 'stage', name: string) {
...
+    // Each file name contains __COMPONENT_SLUG__; its contents contain both
+    // __COMPONENT_NAME__ (PascalCase) and __COMPONENT_SLUG__ (kebab-case).
     const templateFiles = yield* fs.readDirectory(templatesDir);
     const createdFiles: string[] = [];
 
     for (const templateFile of templateFiles) {
       const targetFileName = templateFile.replaceAll('__COMPONENT_SLUG__', slug);
       const sourceContent = yield* fs.readFileString(p.join(templatesDir, templateFile));
       const processedContent = sourceContent
-        .replaceAll('__COMPONENT_NAME_PASCAL__', pascalName)
```
SteinGabriel, line 128: This looks like a regression from when this branch was merged into main. `.replaceAll('__COMPONENT_NAME_PASCAL__', pascalName)` needs to be there, unless it was intentionally removed.

tools/cli/src/main.ts
```diff
@@ -1,93 +1,118 @@
...
-      `\nError: component directory already exists: ${err.path}\n` +
-        `  Choose a different name or delete the existing directory first.\n`,
-    ).pipe(Effect.andThen(Effect.die(err))),
-  ),
-  Effect.provide(GithubReleaseLayer),
-  Effect.provide(NodeContext.layer),
-  (effect) => NodeRuntime.runMain(effect, { disableErrorReporting: true }),
-);
+const program =
+  process.argv[2] === 'mcp'
```
SteinGabriel: The check `process.argv[2] === 'mcp'` short-circuits execution before the CLI parses any arguments. As a result, `--help` is silently ignored and the MCP server is started instead. This may be intentional behavior, but it is worth noting.

tools/cli/src/commands/generate.ts
SteinGabriel, line 69: Noticed a naming inconsistency when generating stages vs callbacks. If you scaffold a stage called `MyCustomStage`, the folder/files come out as `mycustomstage.*` instead of `my-custom-stage.*` ,but the equivalent callback (`MyCustomCallback`) is correctly named as `my-custom-callback`.

This happens because `generate.ts` has two separate slug functions: `toKebabCase` (used for callbacks) splits on case boundaries, but `toSlug` (used for stages) only lowercases and replaces non-alphanumeric chars. It works fine for space-separated stage names like "My Login Stage", but fails on PascalCase input since there's nothing for it to split on.

Same two functions with the same behavior already exist on main (generate.ts:18 and generate.ts:30), so looks like a pre-existing bug, not a regression from `cli-mcp`. Flagging it here since it's related to the MCP work, but it could be split into its own ticket if you'd rather not scope-creep this PR with this fix.

SteinGabriel (approved):
Everything looks good and works as expected, other than a pre-existing issue with custom stage naming, which can be a separate ticket.

PR #502, opened 2026-05-21
Login 2

livstowe:


ryanbas21 (approved):

apps/login-app/src/lib/server/redirect/redirect.effects.ts
```diff
@@ -27,6 +27,25 @@ import type { TokenId } from '$server/schemas';
...
+ */
+export function resolveRealmFromUrl(url: URL): string {
+  const realmParam = url.searchParams.get('realm');
+  if (realmParam != null) {
+    const realm = realmParam.replace(/^\/+/, '');
+    if (!realm) return 'root';
+    return VALID_REALM.test(realm) ? realm : env.FR_REALM_PATH || 'root';
+  }
+  return env.FR_REALM_PATH || 'root';
+}
```
cerebrl: We generally reserve `.effect.ts` files for functions that manage effects (aka side-effects). Since this looks like a pure function, can we move this `*.utilities.ts` file?

apps/login-app/src/routes/(app)/+page.svelte
```diff
@@ -63,10 +63,12 @@
...
        * to support temporarily suspended flows like email verification
-       * and to help AM set journey step successUrl
+       * and to help AM set journey step successUrl.
+       * Only add these when present — undefined would be serialized as the
+       * literal string "undefined" by URLSearchParams and corrupt AM's successUrl.
        */
-      query.goto = data.redirectParams?.goto;
-      query.gotoOnFail = data.redirectParams?.gotoOnFail;
+      if (data.redirectParams?.goto) query.goto = data.redirectParams.goto;
+      if (data.redirectParams?.gotoOnFail) query.gotoOnFail = data.redirectParams.gotoOnFail;
```
cerebrl: I'm not a big fan of the conditional with assignment on a single line like this. Can we use braces and multiple line syntax?

apps/login-app/src/lib/server/sessions.ts
cerebrl, line 1: I think ultimately, I'd like to change the name of this file to better represent its contents. I don't think we need to do it now, but it's worth thinking about.

livstowe:
The redirect flow now preserves SAML goto values when AM returns a console default. Redirect tests cover this case.

livstowe:
Addressed the remaining review feedback in `fa6296f`: clarified the Login2 runtime environment examples/documentation and expanded the conditional query-building block in `+page.svelte`. The redirect E2E fix remains in `2faf602`; all 64 Login2 app unit tests pass.

livstowe:
Removed the unused `COOKIE_SECRET` example contract in `6a0901d`. The current Login2 implementation uses the host-bound AM cookie directly and no longer has process-local/BFF cookie signing state.

livstowe:
Removed the unused `FR_OAUTH_PUBLIC_CLIENT`, `FR_OAUTH_SCOPE`, and optional platform-origin examples from the Login2 runtime env examples in `fa29280`. The app configures its Journey client from server-generated discovery data; these env values were dead contract.

livstowe:
Removed the final unused `APP_DOMAIN` runtime export in `e43fcf9`; source grep now confirms the Login2 app only retains environment variables with active consumers.

livstowe:
Aligned `AGENTS.md`, `CONTRIBUTING.md`, `TESTING.md`, and `e2e/README.md` with the actual Login2 runtime contract in `ea7954b`: FR_AM_URL, FR_AM_COOKIE_NAME, FR_REALM_PATH, and FR_AM_WELLKNOWN_URL. Removed stale VITE/OAuth-client documentation.

livstowe:
Added `2320d91` with focused Login2 session/realm safety tests: configured AM-cookie forwarding, host-only cookie handling, root/bravo OAuth and JSON paths, and traversal rejection. The framework app test suite now reports 69 passing tests.

livstowe:
Latest framework CI now has build, verification, Chromatic, and all E2E shards passing; remaining workflow/security checks and reviewer re-approval are still required. The new session tests add 69 passing app tests locally.

livstowe:
Both of cerebrl's requested changes are already in place on this branch: resolveRealmFromUrl moved to redirect.utilities.ts (it's a pure function), and the two goto/gotoOnFail conditionals in +page.svelte now use braced multi-line syntax. Leaving sessions.ts naming as-is per your own note that it's not blocking. Ready for re-review -- this is the one PR every downstream Login2 step (image publish, saas dark merge, RC) is waiting on.

apps/login-app/src/lib/server/redirect/redirect.effects.ts
livstowe: Done: resolveRealmFromUrl now lives in redirect.utilities.ts, which exports only pure functions. Everything left in redirect.effects.ts (storeRedirectParams, readAndClearRedirectCookie, createRedirectContext, validateUrl) writes cookies or calls AM, so it belongs there.

apps/login-app/src/routes/(app)/+page.svelte
livstowe: Done: both conditionals now use braces with the assignment on its own line.

apps/login-app/src/lib/server/sessions.ts
livstowe: Fair point: sessions.ts now spans cookie helpers, realm path resolution, and AM session calls, so the name undersells it. Agreed to defer the rename out of this PR; happy to fold it into a follow-up so this one stays reviewable.

apps/login-app/src/routes/(app)/+layout.server.ts
```diff
@@ -11,22 +11,35 @@ import { error } from '@sveltejs/kit';
...
+  const wellknownUrl = env.FR_AM_WELLKNOWN_URL;
   const idmBaseUrl = env.FR_IDM_URL ?? amUrl;
 
-  if (!amUrl || !realmPath || !wellknown) {
+  if (!amUrl || !wellknownUrl) {
     throw error(
       500,
-      'Login App is not configured. Ensure the required environment variables like FR_AM_URL, FR_REALM_PATH, and FR_AM_WELLKNOWN_URL are set before starting the app.',
+      'Login App is not configured. Ensure the required environment variables like FR_AM_URL and FR_AM_WELLKNOWN_URL are set before starting the app.',
     );
```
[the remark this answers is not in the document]
livstowe: Fixed: the guard message now names all four required variables, and the tests mock FR_REALM_PATH with a leading slash so the normalization regression is covered (see f9a91d8).

livstowe:
@polaris-opus Please review this PR

livstowe:
@ping-gh-polaris @polaris-opus Please review this PR

apps/login-app/src/routes/(app)/+page.svelte
livstowe: Fixed at the current head — both conditionals now use braces and multi-line syntax.

apps/login-app/src/lib/server/redirect/redirect.effects.ts
livstowe: Agreed on the naming — happy to fold this into a follow-up so this PR stays reviewable (same plan as the sessions.ts rename).

apps/login-app/src/lib/server/sessions.ts
livstowe: Also folding into the follow-up PR with the redirect.effects move — one naming-cleanup PR covering both.

livstowe: Agreed on the naming concern — followed up in PR #592, which renames this file to `am-session.ts` to reflect its contents (AM session/cookie helpers plus the shared proxy-response helper). The rename will land on this branch via the stacked merge once #592 is approved.

apps/login-app/src/routes/(app)/+layout.server.ts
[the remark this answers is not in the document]
livstowe: Addressed in a later commit: the 500 error message now lists all four required variables (FR_AM_URL, FR_AM_COOKIE_NAME, FR_REALM_PATH, FR_AM_WELLKNOWN_URL), matching the import-time REQUIRED_ENV check in core/constants.ts, so the operator guidance is consistent at both layers.

apps/login-app/src/lib/server/redirect/redirect.utilities.ts
```diff
@@ -9,9 +9,24 @@
...
+ */
+export function resolveRealmFromUrl(url: URL): string {
+  const realmParam = url.searchParams.get('realm');
+  if (realmParam != null) {
+    const realm = realmParam.replace(/^\/+/, '');
+    if (!realm) return 'root';
+    return VALID_REALM.test(realm) ? realm : env.FR_REALM_PATH || 'root';
+  }
+  return env.FR_REALM_PATH || 'root';
+}
```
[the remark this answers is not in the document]
livstowe: Applied the changes

apps/login-app/src/routes/(app)/+page.server.ts
```diff
@@ -7,23 +7,75 @@
...
+      const amOrigin = new URL(AM_DOMAIN_PATH).origin;
+      const realm = resolveRealmFromUrl(event.url);
+
+      // Validate the session is live before acting on it.
+      const userId = await getUserIdFromSession(tokenId, realm);
+      if (!userId) throw new Error('Session invalid or expired');
+
+      // If goto is an AM OAuth authorize URL, redirect there directly so the SPA's
+      // iframe PKCE mechanism completes without a full login prompt.
+      const goto = event.url.searchParams.get('goto');
```
cerebrl: Let's make sure we validate this `goto` URL before using it.

apps/login-app/src/lib/server/sessions.ts
```diff
@@ -16,41 +15,61 @@ import type { Cookies } from '@sveltejs/kit';
...
-export function remove(uuid: string): void {
-  amSessions.delete(uuid);
+export function clearAmCookie(cookies: Cookies): void {
+  cookies.delete(AM_COOKIE_NAME, {
+    httpOnly: true,
+    sameSite: 'lax',
+    secure: true,
+    path: '/',
+  });
+}
```
cerebrl: This is non-blocking, but we usually like to avoid these types of "effectful" mutations. We prefer to return new output instead of mutating the given input. It's not a huge deal, but let's just keep this in mind.

apps/login-app/src/routes/api/health/ready/+server.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { RequestHandler } from './$types';
+
+const AM_TIMEOUT_MS = 2000;
+
+function discoveryUrl(): string | null {
+  if (!env.FR_AM_WELLKNOWN_URL || !env.FR_AM_COOKIE_NAME || !env.FR_REALM_PATH) return null;
+  return env.FR_AM_WELLKNOWN_URL;
+}
+
+export const GET: RequestHandler = async () => {
```
cerebrl: One thing we need to think about is tenants that don't have the OAuth2 Provider Service enabled. We discovered this during our SDK Migration in Platform Login. There are plain `forgeops` installations that don't have the same services enabled by default as AIC.

Does that change our implementation here and other places?

apps/login-app/src/routes/api/authenticate/+server.ts
```diff
@@ -1,69 +1,47 @@
...
-
-  const headers = new Headers();
-  headers.append(
-    'set-cookie',
-    cookieUuid
-      ? `cookie=${cookieUuid}; domain=.crbrl.ngrok.io; SameSite=None; HTTPOnly; Secure;`
-      : '',
-  );
+  const setCookie = response.headers.get('set-cookie');
+  if (setCookie) setAmCookie(event.cookies, setCookie);
```
cerebrl: This does look clean, but I do want to avoid these kind of hidden mutations if we can. I'd rather the function return a cookie to be set or a new set of cookies to overwrite the existing cookies.

cerebrl (changes requested):
This looks good. I left a few comments. The only one that's blocking is the `goto` validation.

PR #516, opened 2026-05-28
feat(admin-registration): add AdminRegistration stage for PingOne AIC admin invite flow

vatsalparikh:
## JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-5055

## Description

- Adds a dedicated AdminRegistration stage that handles the PingOne AIC admin invite journey, replacing the generic fallback rendering for these screens
- Stage covers four sub-stages driven by AM callback signals: welcome (with tenant name injection), OTP verification, privacy policy acceptance, and invalid invite

## How to test

- Only use 8443 port because oauth client and CORS only work with this port
- Use the following env vars
```
FR_AM_JOURNEY_LOGIN='Login'
FR_AM_JOURNEY_REGISTER=Register
FR_OAUTH_SCOPE=openid profile me.read
DEVELOPMENT=true
FR_AM_URL=https://openam-vatsal-sdks-test.forgeblocks.com/am
FR_AM_COOKIE_NAME=8fee3e37ffefd83
FR_AM_WELLKNOWN_URL=https://openam-vatsal-sdks-test.forgeblocks.com/am/oauth2/.well-known/openid-configuration
FR_OAUTH_PUBLIC_CLIENT=clientoauth
FR_REALM_PATH=/
```
- Connect with me and I can send an admin registration invite.
- When you receive an admin registration invite, change the domain in the admin registration from `https://openam-vatsal-sdks-test.forgeblocks.com` to `https://localhost:8443`. The query params should stay the same.
- Start the login app locally and load the admin invite URL with query params and follow the admin registration process.
- You can see the invalid invite page if you remove the uuid query param or make the uuid param blank

## Recording

https://github.com/user-attachments/assets/17cfc888-ac9a-45e9-b45c-b409d465fe3c


apps/login-app/src/routes/(app)/+page.svelte
```diff
@@ -65,6 +66,7 @@
        */
       query.goto = data.redirectParams?.goto;
       query.gotoOnFail = data.redirectParams?.gotoOnFail;
+      query.uuid = uuidParam as string;
```
SteinGabriel: I believe `uuidParam` can be `null`/`undefined` here when not passed into as search param.

core/journey/stages/admin-registration.svelte
```diff
@@ -0,0 +1,308 @@
...
+    'Canada',
+    'European Union',
+    'Hong Kong',
+    'Indonesia',
+    'New Zealand',
+    'Singapore',
+    'United Kingdom',
+    'United States',
+    'Rest of the World',
+  ];
```
SteinGabriel: Do we want this list to be hard-coded? Could we fetch this list from somewhere instead?

core/journey/stages/admin-registration.svelte
```diff
@@ -0,0 +1,308 @@
...
+    <Button
+      busy={journey?.loading}
+      classes={otpComplete ? '' : 'tw_opacity-50 tw_pointer-events-none'}
+      style="primary"
+      type="button"
+      width="full"
+      onClick={() => {
+        syncOtpAnswer();
+        form?.submit();
+      }}
```
SteinGabriel: I'm seeing this behavior when the verification code is invalid. I think it'd be nice to have an error message here.
 
https://github.com/user-attachments/assets/c35206ac-751e-43f0-8564-7c3692c10dcf

SteinGabriel (commented):
Initial review.

core/journey/stages/admin-registration.svelte
vatsalparikh: These regions are hard-coded in the scripted node response, so we aren't receiving these from the backend, they are part of the UI script. To replace them, I had to hard code them in the svelte file.

apps/login-app/src/routes/(app)/+page.svelte
vatsalparikh: Yes, if I don't cast it as string, then I have to add a if(query.uuid) check, and it felt unnecessary. So I am simply casting it as string. If the string is empty query.uuid is empty as well, so it's not really an issue.

core/journey/stages/admin-registration.svelte
vatsalparikh: I completely forgot about the invalid code case! Thank you for showing this! I've added another screen for otp retry with its own story and mock, and it should now show an error when the otp is incorrect. Thank you 👍

core/journey/stages/admin-registration.svelte
```diff
@@ -0,0 +1,332 @@
...
+    const hiddenCbs = step.getCallbacksOfType(
+      callbackType.HiddenValueCallback,
+    ) as HiddenValueCallback[];
+
+    const jurisdictionCb = hiddenCbs.find((cb) =>
+      (cb.getOutputByName('id', '') as string).startsWith('jurisdiction-input-'),
+    );
+    const otpCb = hiddenCbs.find(
+      (cb) => (cb.getOutputByName('id', '') as string) === 'p1aic-otp-answer',
+    );
```
ryanbas21: Since we only need one of these it seems from below, let's compute this more intelligently in the sense that we cant compute jurisdictioncb first, and assign it if it's defined, otherwise we can get otpCb.

core/journey/stages/admin-registration.svelte
```diff
@@ -0,0 +1,332 @@
...
+      // Distinguish welcome vs invalid-invite by inspecting the script message
+      const textOutputCbs = step.getCallbacksOfType(
+        callbackType.TextOutputCallback,
+      ) as TextOutputCallback[];
+      const scriptCb = textOutputCbs.find((cb) => cb.getMessageType() === '4');
+      if (scriptCb?.getMessage().includes('Invitation not valid')) {
+        subStage = 'invalidInvite';
+      } else {
+        const msg = scriptCb?.getMessage() ?? '';
+        tenantName = msg.match(/<span[^>]*p1aic-tenant-name[^>]*>([^<]+)</)?.[1] ?? '';
```
ryanbas21: how safe are these hard coded regex patterns across tenant options? like are we expecting this to always be there in the cases where we are checking them?

core/journey/stages/admin-registration.svelte
```diff
@@ -0,0 +1,332 @@
...
+  function handleOtpKeyup(event: KeyboardEvent, index: number) {
+    const input = event.target as HTMLInputElement;
+    if (/^\d$/.test(event.key) && input.value.length === 1) {
+      if (index < 5) otpInputs[index + 1]?.focus();
+      else otpInputs[index]?.blur();
+    } else if (event.key === 'Backspace' && index > 0 && input.value === '') {
+      otpInputs[index - 1]?.focus();
+    }
+    syncOtpAnswer();
+  }
```
ryanbas21: should we break these functions out? this is super imperative to read and i'm wondering how we can structure some of this so it's a bit less dense when reading the logic.

`handleOtpKeyup` needs 
- event handling
- dom elements list
- index checking on that list
- blurring a field/focus a field
- key checking / event handling
- syncing the answer


for a 10 line or so function this is dense. it's intermixing state and effects in a way that I think we should spend time evaluating and seeing how we can simplify.

Here is what i'm thinking:

The function has a few responsibilities that maybe we can break down and compose.

- decision of what to do with the event
- the effectful blur/focus
- state syncronization.

I think this can be broken into something that is more declarative in nature 

- read event
- classify what to do with event (function)
- apply some effect (maybe in classify above)
- sync state.

3/4 functions that may help read a bit more the intent and break out testable blocks of code. This doesn't necesarily need to read like this, but for some north star of where my brain goes

```
function handleOtp(event) {
  return pipe(
   event,
   classifyOtpKey,
   handleOtpKey
);
```
We can move the `syncOtpAnswer` into a reactive block maybe so that we update the setInputValue when the value changes?

```
  $: { 
    hiddenValueCb?.setInputValue(otpDigits.join(''));
}
```

core/journey/stages/admin-registration.svelte
```diff
@@ -0,0 +1,332 @@
...
+    'Hong Kong',
+    'Indonesia',
+    'New Zealand',
+    'Singapore',
+    'United Kingdom',
+    'United States',
+    'Rest of the World',
+  ];
+
+  function onJurisdictionChange(event: Event) {
```
ryanbas21: I think we can make this a reactive block in svelte as well.

ryanbas21 (changes requested):
Here are my initial thoughts to ruminate on. My goal is that this component is super dense with logic and I want to see if we can decompose it into something a bit more fluid

core/journey/stages/admin-registration.svelte
vatsalparikh: I have tested this by creating my own tenant, with open-sdks tenant, and with tenants that Liv created earlier.

However, these are all internal, and it would be good to test these otherwise. So that's a good question actually. I will connect with someone from Platform UI and see if they know more about this. Thanks!

core/journey/stages/admin-registration.svelte
vatsalparikh: Thank you for the feedback! This function does not exist anymore. The file is much more cleaner and leaner now!

core/journey/stages/admin-registration.svelte
vatsalparikh: Thank you for the detailed review and suggestions! Based on your suggestions I have refactored the entire admin-registration.svelte file into multiple small files, made otp related logic functional and declarative, and removed some unnecessary functions like syncOtpAnswer. The entire admin-registration folder and files under it are leaner and cleaner now.

Appreciate the feedback 👍

core/journey/stages/admin-registration.svelte
vatsalparikh: I want to keep this very specific for a reason. We can have multiple hidden callbacks in any stage or response from AM. Keeping it specific helps in narrowing better. Especially in cases where we need to decided which sub-stage to show, it helps to keep things deterministic. If we remove those specific checks there is a chance of false positive where otpCallback is triggered even when it is not.

core/journey/stages/admin-registration.svelte
vatsalparikh: Thanks for the input. I started a conversation with people from Platform UI team and will update this PR and the ticket depending on what they say!

apps/login-app/src/lib/stages/admin-registration/admin-registration.svelte
```diff
@@ -0,0 +1,109 @@
...
+  $: {
+    const hiddenCbs = step.getCallbacksOfType(
+      callbackType.HiddenValueCallback,
+    ) as HiddenValueCallback[];
+    const msg =
+      (step.getCallbacksOfType(callbackType.TextOutputCallback) as TextOutputCallback[])
+        .find((cb) => cb.getMessageType() === '4')
+        ?.getMessage() ?? '';
+
+    const subStageRules: { subStage: SubStage; match: boolean }[] = [
```
ryanbas21: Does this need to be in a reactive block? I think it's static so it's not reactive right?

core/journey/stages/admin-registration/otp.svelte
```diff
@@ -0,0 +1,163 @@
...
+    if (action === 'advance') {
+      otpInputs[index + 1].focus();
+    } else if (action === 'finish') {
+      otpInputs[index].blur();
+    } else if (action === 'retreat') {
+      otpInputs[index - 1].focus();
+    }
+  }
+
+  function extractDigits(clipboardData: DataTransfer | null): string[] {
```
ryanbas21: probably doesnt' need to be in the reactive block either

ryanbas21 (approved):
looks good left a few small nb comments

core/journey/stages/admin-registration/otp.svelte
vatsalparikh: It's already outside the reactive block. 

I moved the reactive blocks after the function definitions and am following repo conventions, so it's easy to read and distinguish whether a function is inside or outside a reactive block!

apps/login-app/src/lib/stages/admin-registration/admin-registration.svelte
vatsalparikh: Right, that makes sense. Moved the relevant pieces outside the reactive block, thanks!

core/journey/stages/admin-registration.svelte
SteinGabriel: Looks great now. Thank you!

themes/default/primitives.cjs
```diff
@@ -407,12 +407,9 @@ module.exports = function (theme) {
       },
       // Double class to increase specificity by 1 level
       '&&[aria-invalid="true"]:focus': {
-        '--outline-color': 'hsl(var(--tw-colors-error-dark-hs),var(--tw-colors-error-dark-l), 0.3)',
         '--bg-color':
           'hsl(var(--tw-colors-background-light-hs),var(--tw-colors-background-light-l))',
-        outlineColor: `var(--outline-color, ${colorLib(theme('colors.error.dark'))
-          .fade(0.3)
-          .toString()})`,
```
SteinGabriel, line 415: What's the reason for removing the outline color/width from the focus state? 
I think this could mean an accessibility regression.

apps/login-app/src/lib/stages/admin-invite-verify-code.svelte
```diff
@@ -0,0 +1,163 @@
...
+    moveFocus(decideFocus(event.key, index), index);
+  }
+
+  function handleOtpPaste(event: ClipboardEvent) {
+    event.preventDefault();
+    otpDigits = extractDigits(event.clipboardData);
+    const nextEmpty = otpDigits.findIndex((digit) => digit === '');
+    if (nextEmpty >= 0) {
+      otpInputs[nextEmpty].focus();
+    }
```
SteinGabriel, line 77: When all code digits are pasted, the focus goes back to the first input instead of the last, so the behaviour of pressing Tab to navigate to the "Next" button is not consistent with typing the digits in manually.

SteinGabriel (changes requested):

themes/default/primitives.cjs
vatsalparikh: I tried getting rid of the border because the border was incomplete and not filling the entire box as shown in the screenshot below. However, this is also happening in the normal case. This is unrelated to the custom stages and is a password callback / stage thing. I will look into this deeper to figure out the best solution.

The change in primitives.cjs simply eliminated the border so the highlight still happens without any dark color. Will work on this to find a better solution, thanks!

Screenshot 2026-06-02 at 5 09 06 PM

apps/login-app/src/lib/stages/admin-invite-verify-code.svelte
vatsalparikh: Very interesting and subtle, I did not realize, thank you!

I've updated the logic to focus on the last element when digits are pasted. Focus on last digit also happens when you enter digits manually, so it is consistent and tab-able to the 'Next' button!

themes/default/primitives.cjs
vatsalparikh: Thank you. I found the root cause and made the fix by adding a class to floating-label, and it only affects the password callback because only password callback has the right icon of "eye". So it should not affect anything globally. Below is the updated screenshot, thanks!
Screenshot 2026-06-03 at 8 49 44 AM

SteinGabriel (approved):
Looks good and works as expected. Thank you!

core/journey/_utilities/map-stage.utilities.ts
```diff
@@ -120,7 +112,26 @@ export function mapStepToStage(currentStep: StepTypes): StageTypes | Component {
...
+      (cb) =>
+        cb.getMessageType() === '4' &&
+        (cb.getMessage().includes('Invitation not valid') ||
+          cb.getMessage().includes('p1aic-tenant-name')),
+    ) ||
+    hiddenValueCallbacks.some((cb) => {
+      const id = cb.getOutputByName('id', '') as string;
+      return id === 'p1aic-otp-answer' || id.startsWith('jurisdiction-input-');
+    });
+  if (isAdminRegistration) return stages['AdminRegistration'] ?? Generic;
```
cerebrl: I think we're conflating two things here: a journey and a stage. `AdminRegistration` is a journey, not a stage. Stages are the independent steps of a journey. A journey can have 0 to _n_ stages. Also, a stage cannot have sub-stages.

With that said, I'm sure you made this decision for a reason, so if you'd like to share that logic, please do. But, I instinctively hesitate here because it's creating a new definition and design that didn't exist before and deviates from our usual taxonomy.

core/journey/stages/step.mock.ts
```diff
@@ -915,6 +917,88 @@ export const successMessagesRenderingStep = {
   ],
 };
 
+export const adminRegWelcomeStep: Step = {
```
cerebrl: I'd like to colocate this mock data with the respective stages.

core/journey/journey.interfaces.ts
```diff
@@ -62,6 +63,14 @@ export interface StageJourneyObject {
   redirect: (step: JourneyStep) => Promise<void>;
   stack: StackStore;
 }
+export type StageComponent = Component<{
```
cerebrl: This type doesn't really define a "Stage Component", pre se, but describes a "Journey Component". Can we rename this?

experimental/custom/demo/stages/custom-login/custom-login.svelte
```diff
@@ -126,24 +126,17 @@
...
-      <!--
-        Logo / icon placeholder.
-        Replace this <div> with your own logo, e.g.:
-          <img src="/your-logo.svg" alt="Acme Corp" class="tw_h-12" />
-        Or import and use a Svelte icon component from your design system.
-      -->
-      <div
-        class="tw_w-16 tw_h-16 tw_rounded-full tw_bg-blue-600 tw_flex tw_items-center tw_justify-center"
-        aria-hidden="true"
+    <div class="tw_flex tw_items-baseline tw_gap-3 tw_mb-4">
```
cerebrl: Can you explain the motivation behind this change? I'm not sure I understand its connection to your story.

core/journey/journey.svelte
```diff
@@ -18,12 +18,13 @@
   import { mapStepToStage } from '$journey/_utilities/map-stage.utilities';
   import { stack } from '$journey/journey.store';
 
-  import type { JourneyStore } from '$journey/journey.interfaces';
+  import type { JourneyStore, StageComponent } from '$journey/journey.interfaces';
```
cerebrl: Yeah, I think `JourneyComponent` is a more appropriate name here, rather than `StageComponent`.

apps/login-app/src/lib/stages/index.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ **/
+
+import AdminRegistration from './admin-registration/admin-registration.svelte';
+import MfaEnrollment from './mfa-enrollment.svelte';
+
+import type { StageComponent } from '$journey/journey.interfaces';
+
+export const appStages: Record<string, StageComponent> = {
+  AdminRegistration,
```
cerebrl: I don't have any hesitation with grouping stages together in a directory, but I'm not sure about creating a hierarchical stage with sub-stages. I'd like to see this "flattened out", so the sub-stages are converted to top-level stages.

cerebrl (changes requested):

core/components/compositions/input-floating/floating-label.svelte
```diff
@@ -50,7 +50,7 @@
     {forceValidityFailure}
     {isFirstInvalidInput}
     inputClasses={`tw_input-floating dark:tw_input-floating_dark ${
-      hasRightIcon ? '!tw_border-r-0 !tw_rounded-r-none' : ''
+      hasRightIcon ? '!tw_border-r-0 !tw_rounded-r-none tw_input-has-button' : ''
```
cerebrl, line 53: What is this addressing?

core/journey/_utilities/map-stage.utilities.ts
vatsalparikh: This is an aha moment for me. I never thought of admin registration and mfa as journeys because they behave as **sub-journeys**. They are not full-blown journeys onto themselves and act like transitions (meaning journey within a journey and we do have that concept in admin console UI where we can define journey within a journey). I will go back to the drawing board (metaphorically) and re-organize the code around this idea of admin registration and mfa as journeys, and I agree that those sub-stages should be defined as stages, thank you!

core/components/compositions/input-floating/floating-label.svelte
vatsalparikh: `tw_input-has-button` class ensures the focus ring renders on all four sides when the input has an adjacent button, without it the right edge of the ring gets clipped behind the button. Shared screenshots below.

Screenshot 2026-06-10 at 5 16 06 PM
Screenshot 2026-06-10 at 5 16 27 PM

core/journey/stages/step.mock.ts
vatsalparikh: This will introduce inverse dependency on login app. Both map-stage.utilities.test file in core and stories tests in login app need them. We can make copies of them (one for core and one for login app) but I don't like the idea of duplicating the same data. Let me know what you think.

apps/login-app/src/lib/stages/index.ts
vatsalparikh: Thanks, flattened! No more sub-stages!

core/journey/journey.svelte
vatsalparikh: Continued to use `StageComponent` now that I am exporting stages, not journeys!

core/journey/journey.interfaces.ts
vatsalparikh: Continued to use `StageComponent` now that I am exporting stages, not journeys!

experimental/custom/demo/stages/custom-login/custom-login.svelte
vatsalparikh: Chromatic found regressions / visual changes for this demo component. I've restored the demo custom login file. It's a demo file anyway, so i have restored the file for now.

core/journey/stages/step.mock.ts
vatsalparikh: I thought about this further. I've moved the detection logic to login-app as well, so now the detection along with stages are all defined in login-app. This naturally moves all the testing responsibility to login-app. map-stage only calls the detect function already defined in login-app. All tests, mocks for login-app specific stages are now moved to login-app. Clean!

core/journey/journey.svelte
```diff
@@ -18,12 +18,13 @@
...
   import { stack } from '$journey/journey.store';
 
-  import type { JourneyStore } from '$journey/journey.interfaces';
+  import type { JourneyStore, StageRegistryEntry } from '$journey/journey.interfaces';
 
   export let componentStyle: 'app' | 'inline' | 'modal';
   export let displayIcon: boolean;
   export let formEl: HTMLFormElement | null = null;
   export let journeyStore: JourneyStore;
+  export let stages: Record<string, StageRegistryEntry> = {};
```
cerebrl: I really like the way this is implemented now. The only change I'd like to see is a different name. `stages` is just too generic for me. Can we change it to `externalStages` or something? That will help other trace the origin of these additional stage components.

cerebrl (approved):
I'm approving this PR, but I have one request before merge.

core/journey/journey.svelte
vatsalparikh: Makes sense! I named those stages as `loginAppStages` when we are defining them in login app, and as soon as we pass them to journey as a prop, the prop is named as `externalStages`. This way we know that the prop is external and that it is coming from login app.

Done, thanks!

PR #520, opened 2026-06-04
SDKS-5079: Add IDM Theming Support to Widget and Login App

SteinGabriel:
## Summary

https://pingidentity.atlassian.net/browse/SDKS-5079

Adds IDM-driven theming to the Login Widget and Login App. Consumers pass design tokens via
`await configure({ style: { theme } })`. The Login App SSR layer fetches the active IDM theme
from `/openidm/config/ui/themerealm` on each page load and applies it as CSS custom properties
before the initial HTML is sent, so there is no flash of unstyled content.

### `core/style.store.ts`

- `themeSchema` (Zod): validates all supported theme fields. Hex fields use a strict 6/8-digit
  regex with `.catch(undefined)` so an invalid value degrades to unset instead of discarding the
  whole theme. `logo` / `favicon` are URL-validated, `fontFamily` is regex-validated (CSS
  injection guard).
- `ThemeObject` type export. Field names match the internal CSS var they target
  (`buttonBorderRadius` -> `--fr-button-border-radius`).
- `styleSchema` extended with `theme: themeSchema.optional()`.
- `urlRegex` exported for reuse in `idm-theme.effects.ts`.

### `core/_utilities/theme.utilities.ts` _(new)_

- `hexToHslChannels(hex)` -> `{ hs: "H, S%", l: "L%" }`. Accepts 6- or 8-digit hex (alpha
  stripped), throws on anything else. Channels stay split because component classes reference
  them independently, e.g. `calc(var(--tw-colors-*-l) - 10%)`.
- `buildThemeVarsEntries(theme)` -> `[string, string][]` of CSS var declarations. Pure mapping
  core shared by the SSR path (`+layout.svelte`) and the DOM path (`applyThemeVars`). Invalid
  hex per field is skipped silently.
- `encodeCssUrl(url)` -> `url("…")` with `"` percent-encoded, preventing breakout from a quoted
  `url()` inside an inline `style` attribute.

Notable mappings: `primaryColor` fills both primary-dark and primary-light slot pairs;
`secondaryColor` fills all three secondary pairs (dark/default/light); `backgroundColor` fills
the background pairs plus `--fr-page-bg-color`; `inputFocusRingColor` sets only
`--fr-focus-ring-color`, deliberately not the shared `--tw-colors-focus-default-*`, so button and
link focus transitions are unaffected.

### `core/_effects/theme.effects.ts` _(new)_

- `applyThemeVars(rootEl, theme)` — DOM side-effect wrapper over `buildThemeVarsEntries`; noop
  when either argument is absent. Used by the widget standalone path.

### `packages/login-widget/src/lib/index.svelte`

- `bind:this={widgetRootEl}` on both modal and inline roots.
- `$: applyThemeVars(widgetRootEl, $styleStore?.theme)` reapplies on every store update.

### `packages/login-widget/src/lib/custom.base.css`

- `:where(.fr_widget-root)` gains `color: var(--fr-body-text-color, inherit)` and
  `var(--fr-font-family, 'Open Sans')` at the head of the font stack.

### `themes/default/token-vars.cjs` _(new)_

- `buildTokenVars(theme)` — Tailwind plugin helper. Resolves 25 token paths through Tailwind's
  `theme()` resolver, converts each hex to an HSL channel pair, and returns a
  `Record<string, string>` of `--tw-colors-*-hs` / `-l` vars.

### `themes/default/config.cjs`

- Plugin merges `':where(:root)': buildTokenVars(theme)` into `addComponents`, covering both the
  login-app and widget standalone builds. Specificity 0, so any consumer `:root {}` or IDM inline
  style wins.

### `themes/default/tokens.cjs`

- New `colors.primary.off` token (`slate.700`), target of `primaryOffColor`.

### `themes/default/primitives.cjs`

New CSS var slots: `--fr-button-border-radius` on `.button-base`; `--fr-input-text-color` on
`.input-base` / `_dark`; `--fr-button-text-color` on `.button-primary`; `--fr-input-bg-color` on
`.input-base`, `.select-base`, `.password-button` and their dark and hover/focus states;
`--fr-input-border` / `--fr-input-border-color-value` on input borders; `--fr-input-label-color`
on `.input-label` / `_dark`; `--fr-button-focus-ring-color` on `.button-base:focus`, isolated
from the input focus ring; `--fr-focus-ring-color` on input and select `:focus`;
`--fr-select-accent-color` / `--fr-select-hover-bg-color` on select states. The 3px focus ring is
restored on `.focusable-element:focus`; the 1px ring is scoped to input and select only.

### `themes/default/boxes.cjs`, `compositions.cjs`, `utilities.cjs`, `journey.cjs`

- `--fr-card-bg-color` on `.dialog-box`, `.containing-box` and dark variants.
- `--fr-card-border-radius` on `.dialog-box`, `.dialog-header` (top corners), `.dialog-x`,
  `.containing-box`.
- `--fr-card-text-color` with `inherit` fallback on `.dialog-body`.
- `--fr-focus-ring-color`, `--fr-select-accent-color`, `--fr-select-hover-bg-color` on
  `.focusable-element:focus`, `accent-color`, select hover.
- `--fr-input-bg-color` fallback wrapper around select `--bg-color` in light and dark.

### `apps/login-app/src/lib/server/idm-theme.effects.ts` _(new)_

- `fetchIdmTheme(idmBaseUrl, realmPath, journeyName)` — anonymous GET of
  `/openidm/config/ui/themerealm`. Resolves the active theme by `?journey=` -> `linkedTrees` ->
  `isDefault` -> first entry. Validates through `themeSchema.safeParse()`. Returns `undefined`
  theme on any failure. Default 1500ms timeout via `AbortSignal.timeout`, overridable with
  `FR_IDM_THEME_TIMEOUT_MS`. Module-level cache keyed `realm::journey`; every failure path falls
  back to the last successful result before falling back to widget defaults.
- `toIdmTheme()` — translates IDM external field names (`buttonRounded`) to internal
  `ThemeObject` names (`buttonBorderRadius`); drops invalid logo/favicon URLs individually.
- Returns `{ theme, backgroundImageUrl }`.

### `apps/login-app/src/routes/(app)/+layout.server.ts`

- `idmBaseUrl = env.FR_IDM_URL ?? env.FR_AM_URL` — IDM can be pointed at a separate host.
- `journeyName = url.searchParams.get('journey') ?? env.FR_AM_JOURNEY_LOGIN ?? null`.
- Returns `idmTheme` and `backgroundImageUrl`; all IDM logic stays in `idm-theme.effects.ts`.

### `apps/login-app/src/routes/(app)/+layout.ts`

- Calls `initializeStyles({ theme: data.idmTheme, logo: { light, dark, height } })` when
  `idmTheme` is defined; `logo` is only spread in when `idmTheme.logo` is present.

### `apps/login-app/src/routes/(app)/+layout.svelte`

- Theme vars serialized into an inline `style` attribute on `<div class="theme-root">`, present
  in the initial HTML. No `{@html}`; Svelte escapes the attribute value.
- `--fr-page-bg-image` set from `data.backgroundImageUrl` in the same attribute.
- `<link rel="icon" href={data.idmTheme?.favicon ?? '/favicon.ico'} />`.

### `core/journey/stages/generic.svelte`, `login.svelte`

- Logo rendered from `$styleStore.logo` at the IDM-specified height (default 72x200), guarded
  with `componentStyle !== 'modal'` so it does not double-render in modal mode. URLs pass through
  `encodeCssUrl()`.

### `core/components/compositions/dialog/dialog.svelte`

- Existing logo `url()` interpolations replaced with `encodeCssUrl()`.

### `core/components/primitives/box/centered.svelte`

- `.page-shell` reads `--fr-page-bg-color` with an `hsl(var(--tw-colors-body-light-*))` fallback
  (body-dark under `.dark` and `prefers-color-scheme: dark`), plus `--fr-page-bg-image` and
  `--fr-font-family`.

### Config / tooling

- `.npmrc`: `public-hoist-pattern[]=@effect/*`, `public-hoist-pattern[]=effect`.
- `.env.example`, `.env.docker.example`: document optional `FR_AM_JOURNEY_LOGIN`.
- `core/style.stories.js` / `style.story.svelte`: `ThemeOverride` story under `Style/Config`.
- `apps/login-app/src/routes/e2e/widget/{inline,modal}/theme/+page.svelte`: E2E fixtures.
- Changeset: `@forgerock/login-widget` minor.

  ## Tests

  | File | Tests | Coverage |
  |------|-------|----------|
  | `core/_utilities/theme.utilities.test.ts` _(new)_ | 18 | `hexToHslChannels`,
  `buildThemeVarsEntries`, `encodeCssUrl` |
  | `core/_effects/theme.effects.test.ts` _(new)_ | 27 | `applyThemeVars` DOM application,
  all field groups |
  | `core/style.store.test.ts` _(new)_ | 23 | `themeSchema` validation, `styleSchema`
  extension, `initialize()` integration |
  | `apps/login-app/src/lib/server/idm-theme.effects.test.ts` _(new)_ | 15 | `fetchIdmTheme`
  resolution logic, cache fallback, graceful degradation |
  | `e2e/tests/widget/modal/widget-modal.theme.test.js` _(new)_ | 4 | CSS vars applied from
  API; no vars on default; partial theme; invalid hex ignored |

  **245/245 unit tests pass · 4/4 E2E theme tests pass**

  ## How to test

  ### 1. Widget theme API (no server required)

  ```js
  import { configuration } from '@forgerock/login-widget';

  configuration({
    style: {
      theme: {
        primaryColor: '#cc0000',
        buttonBorderRadius: 20,
        cardBorderRadius: 16,
      },
    },
  });

  Inspect .fr_widget-root in DevTools — --tw-colors-primary-dark-hs,
  --tw-colors-primary-dark-l, --fr-button-border-radius, --fr-card-border-radius should all
  be set.

  2. Storybook visual check

  pnpm storybook

  Open Style › ThemeOverride — button red, card corners rounded. Open Style › Default — no
  inline CSS vars, no visual change.

  3. Login App SSR integration

  Set FR_AM_URL to an AM instance with IDM co-located. Set FR_AM_JOURNEY_LOGIN=Login. Load
  /?journey=Login. The login card should render using the IDM theme for the Login tree.
  Confirm --fr-card-bg-color, --fr-input-bg-color, logo, and primaryColor match IDM values
  and are present in the initial HTML (view source — not just DevTools).

  4. Fallback / negative path

  Point FR_AM_URL at an unreachable host. Page should load with widget defaults — no error
  thrown, no broken layout.

  5. Run unit and E2E tests

  pnpm --filter @forgerock/login-widget exec vitest run 'theme'
  pnpm --filter @forgerock/login-widget exec vitest run 'style.store'
  pnpm --filter @forgerock/login-app test
  pnpm build:app && pnpm ci:e2e -- tests/widget/modal/widget-modal.theme.test.js

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -5,6 +5,7 @@ FR_REALM_PATH=<<<(string) Realm path of AM>>>
 FR_OAUTH_PUBLIC_CLIENT=<<<(string) Your Public OAuth client name/ID>>>
 # FR_OAUTH_SCOPE=openid profile me.read  # optional — AM uses client defaults if omitted
 FR_AM_WELLKNOWN_URL=<<<(string) Your OIDC Discovery Wellknown Endpoint>>>
+# FR_AM_JOURNEY_LOGIN=Login  # optional — default login journey name; used for IDM theme resolution when no ?journey= param is present
 
 # BFF session signing
 COOKIE_SECRET=<<<(string) At least 32 characters for HMAC signing>>>
--- a/apps/login-app/src/routes/(app)/+layout.server.ts
+++ b/apps/login-app/src/routes/(app)/+layout.server.ts
@@ -10,15 +10,17 @@
 import { error } from '@sveltejs/kit';
 
 import { env } from '$env/dynamic/private';
+import { fetchIdmTheme } from '$server/idm-theme.effects';
 
 import type { LayoutServerLoad } from './$types';
 
-export const load: LayoutServerLoad = () => {
+export const load: LayoutServerLoad = async ({ url }) => {
   const amUrl = env.FR_AM_URL;
   const clientId = env.FR_OAUTH_PUBLIC_CLIENT;
   const realmPath = env.FR_REALM_PATH;
   const wellknown = env.FR_AM_WELLKNOWN_URL;
   const scope = env.FR_OAUTH_SCOPE;
+  const idmBaseUrl = env.FR_IDM_URL ?? amUrl;
 
   if (!amUrl || !realmPath || !wellknown) {
     throw error(
@@ -27,9 +29,16 @@ export const load: LayoutServerLoad = () => {
     );
   }
 
+  const journeyName = url.searchParams.get('journey') ?? env.FR_AM_JOURNEY_LOGIN ?? null;
+  const { theme: idmTheme, backgroundImageUrl } = idmBaseUrl
+    ? await fetchIdmTheme(idmBaseUrl, realmPath, journeyName)
+    : { theme: undefined, backgroundImageUrl: undefined };
+
   return {
     amUrl,
+    backgroundImageUrl,
     clientId,
+    idmTheme,
     realmPath,
     scope,
     wellknown,
--- a/apps/login-app/src/routes/(app)/+layout.svelte
+++ b/apps/login-app/src/routes/(app)/+layout.svelte
@@ -1,11 +1,28 @@
 <!--
- 
+
  Copyright © 2025 - 2026 Ping Identity Corporation. All right reserved.
- 
+
  This software may be modified and distributed under the terms
  of the MIT license. See the LICENSE file for details.
- 
+
  -->
+
+<script lang="ts">
+  import { buildThemeVarsEntries } from '$core/_utilities/theme.utilities';
+
+  import type { ThemeObject } from '$core/style.store';
+
+  export let data: { idmTheme?: ThemeObject; backgroundImageUrl?: string };
+
+  $: themeStyle = [
+    ...(data.idmTheme ? buildThemeVarsEntries(data.idmTheme) : []),
+    ...(data.backgroundImageUrl
+      ? [['--fr-page-bg-image', `url("${data.backgroundImageUrl}")`] as [string, string]]
+      : []),
+  ]
+    .map(([k, v]) => `${k}:${v}`)
+    .join(';');
+</script>
 
 <svelte:head>
   <!--
@@ -21,7 +38,7 @@
 
   <meta charset="utf-8" />
   <title>Login Application</title>
-  <link rel="icon" href="/favicon.ico" />
+  <link rel="icon" href={data.idmTheme?.favicon ?? '/favicon.ico'} />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <style>
     /**
@@ -68,6 +85,9 @@
     .root {
       height: 100%;
       min-height: 100%;
+    }
+
+    html {
       background-color: #f6f8fa;
     }
 
@@ -84,4 +104,12 @@
   </style>
 </svelte:head>
 
-<slot />
+<div class="theme-root" style={themeStyle}>
+  <slot />
+</div>
+
+<style>
+  .theme-root {
+    height: 100%;
+  }
+</style>
--- a/apps/login-app/src/routes/e2e/widget/modal/theme/+page.svelte
+++ b/apps/login-app/src/routes/e2e/widget/modal/theme/+page.svelte
@@ -1,0 +1,73 @@
+<!--
+
+ Copyright © 2025-2026 Ping Identity Corporation. All right reserved.
+
+ This software may be modified and distributed under the terms
+ of the MIT license. See the LICENSE file for details.
+
+ -->
+
+<script lang="ts">
+  import { onMount } from 'svelte';
+
+  import { page } from '$app/stores';
+  import Widget, { component, configuration, journey } from '$package/index';
+
+  let componentEvents: ReturnType<typeof component> | undefined;
+  let journeyEvents: ReturnType<typeof journey> | undefined;
+  let widgetEl: HTMLDivElement;
+
+  onMount(() => {
+    const params = $page.url.searchParams;
+    const primaryColorParam = params.get('primaryColor');
+    const buttonBorderRadiusParam = params.get('buttonBorderRadius');
+    const cardBorderRadiusParam = params.get('cardBorderRadius');
+
+    configuration({
+      journeyClient: {
+        serverConfig: {
+          wellknown:
+            'https://openam-sdks.forgeblocks.com/am/oauth2/alpha/.well-known/openid-configuration',
+        },
+      },
+      forgerock: {
+        clientId: 'WebOAuthClient',
+        redirectUri: `${window.location.origin}/callback`,
+        scope: 'openid profile email me.read',
+        serverConfig: {
+          baseUrl: 'https://openam-sdks.forgeblocks.com/am/',
+          timeout: 5000,
+        },
+        realmPath: 'alpha',
+      },
+      style: {
+        theme: {
+          ...(primaryColorParam !== null ? { primaryColor: primaryColorParam } : {}),
+          ...(buttonBorderRadiusParam !== null
+            ? { buttonBorderRadius: Number(buttonBorderRadiusParam) }
+            : {}),
+          ...(cardBorderRadiusParam !== null
+            ? { cardBorderRadius: Number(cardBorderRadiusParam) }
+            : {}),
+        },
+      },
+    });
+
+    componentEvents = component();
+    journeyEvents = journey();
+
+    new Widget({ target: widgetEl });
+  });
+</script>
+
+<div class="tw_p-6">
+  <button
+    on:click={() => {
+      journeyEvents?.start({ journey: 'TEST_Login' });
+      componentEvents?.open();
+    }}
+  >
+    Open Login Modal
+  </button>
+</div>
+<div bind:this={widgetEl}></div>
--- a/core/_utilities/theme.utilities.ts
+++ b/core/_utilities/theme.utilities.ts
@@ -1,0 +1,208 @@
+/**
+ *
+ * Copyright © 2025 - 2026 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ **/
+
+import type { ThemeObject } from '$core/style.store';
+
+/**
+ * Builds a safe CSS `url("…")` value from an untrusted URL string.
+ *
+ * The widget injects consumer- and IDM-supplied logo URLs into inline `style`
+ * attributes as `url("<value>")`. A raw `"` in the value would close the quoted
+ * string early and let an attacker append arbitrary CSS declarations. We
+ * percent-encode the only character that can break out of the quoted form (`"`),
+ * which a URL never needs literally. Returns the full `url("…")` token ready to
+ * drop into a declaration.
+ */
+export function encodeCssUrl(url: string): string {
+  return `url("${url.replace(/"/g, '%22')}")`;
+}
+
+/**
+ * Converts a CSS hex color string to HSL channel pair strings matching the
+ * --tw-colors-*-hs / --tw-colors-*-l slot format used in compiled component classes.
+ *
+ * Returns `{ hs: "H, S%", l: "L%" }` — the two values are kept separate because
+ * component classes reference them independently (e.g. `calc(var(--tw-colors-*-l) - 10%)`).
+ *
+ * Accepts 6-digit hex strings (#RRGGBB or RRGGBB). Throws on anything else.
+ */
+export function hexToHslChannels(hex: string): { hs: string; l: string } {
+  let normalized = hex.startsWith('#') ? hex.slice(1) : hex;
+
+  if (/^[0-9a-fA-F]{8}$/.test(normalized)) {
+    normalized = normalized.slice(0, 6);
+  }
+
+  if (!/^[0-9a-fA-F]{6}$/.test(normalized)) {
+    throw new Error(`Invalid hex color: "${hex}"`);
+  }
+
+  const red = parseInt(normalized.slice(0, 2), 16) / 255;
+  const green = parseInt(normalized.slice(2, 4), 16) / 255;
+  const blue = parseInt(normalized.slice(4, 6), 16) / 255;
+
+  const max = Math.max(red, green, blue);
+  const min = Math.min(red, green, blue);
+  const delta = max - min;
+
+  const lightness = (max + min) / 2;
+
+  let hue = 0;
+  if (delta !== 0) {
+    if (max === red) {
+      hue = ((green - blue) / delta) % 6;
+    } else if (max === green) {
+      hue = (blue - red) / delta + 2;
+    } else {
+      hue = (red - green) / delta + 4;
+    }
+    hue = Math.round(hue * 600) / 10;
+    if (hue < 0) hue += 360;
+  }
+
+  const saturation =
+    delta === 0 ? 0 : Math.round((delta / (1 - Math.abs(2 * lightness - 1))) * 1000) / 10;
+
+  const lightnessPercent = Math.round(lightness * 1000) / 10;
+
+  return {
+    hs: `${hue}, ${saturation}%`,
+    l: `${lightnessPercent}%`,
+  };
+}
+
+/**
+ * Maps an IDM theme object to a flat list of CSS custom property name/value
+ * pairs. Invalid or missing fields are silently omitted.
+ */
+export function buildThemeVarsEntries(theme: ThemeObject): [string, string][] {
+  const entries: [string, string][] = [];
+
+  const cssVar = (name: string, value: string): void => {
+    entries.push([name, value]);
+  };
+
+  const hslPair = (hsVar: string, lVar: string, hex: string): void => {
+    try {
+      const { hs, l } = hexToHslChannels(hex);
+      entries.push([hsVar, hs], [lVar, l]);
+    } catch {
+      // invalid hex — skip
+    }
+  };
+
+  const hslCompound = (varName: string, hex: string): void => {
+    try {
+      const { hs, l } = hexToHslChannels(hex);
+      cssVar(varName, `hsl(${hs}, ${l})`);
+    } catch {
+      // invalid hex — skip
+    }
+  };
+
+  const hslSlotWithCompound = (baseName: string, hex: string): void => {
+    try {
+      const { hs, l } = hexToHslChannels(hex);
+      cssVar(`${baseName}-hs`, hs);
+      cssVar(`${baseName}-l`, l);
+      cssVar(baseName, `hsl(var(${baseName}-hs), var(${baseName}-l))`);
+    } catch {
+      // invalid hex — skip
+    }
+  };
+
+  if (theme.primaryColor) {
+    hslPair('--tw-colors-primary-dark-hs', '--tw-colors-primary-dark-l', theme.primaryColor);
+    hslPair('--tw-colors-primary-light-hs', '--tw-colors-primary-light-l', theme.primaryColor);
+  }
+  if (theme.primaryOffColor) {
+    hslPair('--tw-colors-primary-off-hs', '--tw-colors-primary-off-l', theme.primaryOffColor);
+  }
+  if (theme.secondaryColor) {
+    hslPair('--tw-colors-secondary-dark-hs', '--tw-colors-secondary-dark-l', theme.secondaryColor);
+    hslPair(
+      '--tw-colors-secondary-default-hs',
+      '--tw-colors-secondary-default-l',
+      theme.secondaryColor,
+    );
+    hslPair(
+      '--tw-colors-secondary-light-hs',
+      '--tw-colors-secondary-light-l',
+      theme.secondaryColor,
+    );
+  }
+  if (theme.backgroundColor) {
+    hslPair(
+      '--tw-colors-background-dark-hs',
+      '--tw-colors-background-dark-l',
+      theme.backgroundColor,
+    );
+    hslPair(
+      '--tw-colors-background-light-hs',
+      '--tw-colors-background-light-l',
+      theme.backgroundColor,
+    );
+    cssVar('--fr-page-bg-color', theme.backgroundColor);
+  }
+  if (theme.linkColor) {
+    hslPair('--tw-colors-link-dark-hs', '--tw-colors-link-dark-l', theme.linkColor);
+  }
+  if (theme.linkActiveColor) {
+    hslPair('--tw-colors-link-light-hs', '--tw-colors-link-light-l', theme.linkActiveColor);
+  }
+  if (theme.fontFamily) {
+    cssVar('--fr-font-family', theme.fontFamily);
+  }
+  if (theme.buttonBorderRadius !== undefined) {
+    cssVar('--fr-button-border-radius', `${theme.buttonBorderRadius}px`);
+  }
+  if (theme.cardBorderRadius !== undefined) {
+    cssVar('--fr-card-border-radius', `${theme.cardBorderRadius}px`);
+  }
+  if (theme.cardBgColor) {
+    cssVar('--fr-card-bg-color', theme.cardBgColor);
+  }
+  if (theme.inputBgColor) {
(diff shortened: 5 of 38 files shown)


apps/login-app/src/routes/(app)/+layout.server.ts
```diff
@@ -9,16 +9,136 @@
...
+  realmPath: string,
+  journeyName: string | null,
+): Promise<IdmThemeResult> {
+  const url = new URL('/openidm/config/ui/themerealm', idmBaseUrl).href;
+  let response: Response;
+
+  try {
+    response = await fetch(url, {
+      headers: { Accept: 'application/json' },
+      signal: AbortSignal.timeout(3000),
```
vatsalparikh: A 3 second timeout is too long, I believe. Is a 3 second timeout there for a reason? Does the IDM endpoint take this long in some / all cases to return themes? A 3 second timeout means that the page is blocked from rendering for 3 seconds in case of failures.

We should reduce the timeout to about 500 milliiseconds or less ideally.

And I think caching the theme is also a good idea so we can load the last used theme at least. Falling back to default theme is also valid, but I feel that falling back to cache is more accurate.

core/_utilities/theme.utilities.ts
```diff
@@ -0,0 +1,228 @@
...
+    l: `${lightnessPercent}%`,
+  };
+}
+
+/**
+ * Applies IDM theme values as CSS custom properties on a root element.
+ * Group A/B fields set --tw-colors-* HSL slots; Group D fields set --fr-* slots.
+ * Noop when `theme` is undefined or the element reference is null.
+ */
+export function applyThemeVars(rootEl: HTMLElement | null, theme: ThemeObject | undefined): void {
```
vatsalparikh: `applyThemeVars` function is updating the DOM at runtime. I think this is an effectful function and should be moved to an effects file.

themes/default/primitives.cjs
vatsalparikh, line 220: I don't see `primaryOffColor` defined anywhere during compilation / build time. While we are setting the value when we apply the theme that value won't be set because we don't have it defined in default theme. Is that by design or a mistake?

vatsalparikh (commented):
First review

apps/login-app/src/routes/(app)/+layout.server.ts
SteinGabriel: The only reason for the timeout is to prevent the request from getting stuck indefinitely, but I agree 3s was too much for this case. I reduced it to 500ms as you suggested, but even locally with a good internet connection it wasn't enough to fetch the theme the first time, falling back to the login app's default UI. I increased it to 1.5s and that worked fine for me. I added support for configuring it via an `.env` var, in case it's needed. 
I also added a cache fallback system. I think that's a good idea!
Thanks!

core/_utilities/theme.utilities.ts
SteinGabriel: Yeah, good point. I moved `applyThemeVars` to `core/_effects/theme.effects.ts`. `theme.utilities.ts` is now pure. Tests updated accordingly.

themes/default/primitives.cjs
SteinGabriel: `primaryOffColor` is one of the properties that already had a CSS slot (`tw-colors-primary-off-hs` and `tw-colors-primary-off` in `custom.base.css`), so it's mapped and available to be used. It's just not being used by any UI element in login-app or login-widget at the moment, like every other Tailwind class that is available but unused.

apps/login-app/src/routes/(app)/+layout.server.ts
vatsalparikh:
```
curl -w "\nDNS: %{time_namelookup}s\nTTFB: %{time_starttransfer}s\nTotal: %{time_total}s\n" \
  -H "Accept: application/json" \
  "https://openam-sdks.forgeblocks.com/openidm/config/ui/themerealm" \
  -s -o /tmp/idm_theme.json && echo "Status: OK" || echo "Status: error"
cat /tmp/idm_theme.json | head -c 300
```

I ran the curl command to check how fast the endpoint responds and it returned this

```
DNS: 0.002280s
TTFB: 0.089426s
Total: 0.130294s
Status: OK
```

So the actual response is around 130 milliseconds. There is something else that's taking so long, not sure what it is. Also, I did look at /tmp/idm_theme.json, and it returned all themes from alpha and bravo.

vatsalparikh: Scenario           TTFB          Status
Cold start (first request after server start).         1.36s           200 :white_check_mark:
Warm – 2nd request                                               69ms           200 :white_check_mark:
Warm – 3rd request                                                73ms           200 :white_check_mark:
Conclusion: There is no 3-second delay. The cold start is ~1.4 seconds, which is Vite doing SSR module evaluation on the first hit — completely normal in dev mode. Warm requests (which is everything after the first) are a fast ~70ms.

3 second timeout is good. I thought the page is blocked 3 seconds on every page load, but that does not seem to be the case.

apps/login-app/src/routes/(app)/+layout.server.ts
```diff
@@ -9,16 +9,160 @@
...
+ */
+const DEFAULT_IDM_FETCH_TIMEOUT_MS = 1500;
+const themeCache = new Map<string, IdmThemeResult>();
+
+function resolveIdmFetchTimeoutMs(): number {
+  const parsed = Number(env.FR_IDM_THEME_TIMEOUT_MS);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : DEFAULT_IDM_FETCH_TIMEOUT_MS;
+}
+
+async function fetchIdmTheme(
```
cerebrl: Can we move this to its own module? Since this is a top-level file, I'd rather it delegate responsibility, rather than implement it.

apps/login-app/src/routes/(app)/+layout.server.ts
```diff
@@ -9,16 +9,160 @@
...
+    journeyInputBorderColor: matched.journeyInputBorderColor,
+    journeyInputLabelColor: matched.journeyInputLabelColor,
+    journeyInputFocusBorderColor: matched.journeyInputFocusBorderColor,
+    journeyInputSelectColor: matched.journeyInputSelectColor,
+    journeyInputSelectHoverColor: matched.journeyInputSelectHoverColor,
+    buttonFocusBorderColor: matched.buttonFocusBorderColor,
+    journeyInputTextColor: matched.journeyInputTextColor,
+    journeyCardTextColor: matched.journeyCardTextColor,
+    bodyText: matched.bodyText,
+    textColor: matched.textColor,
```
cerebrl: Can we rename the props to which we're mapping the IDM properties? I'd like to preserve the internal naming conventions within the project, and just have a translation layer that takes in external data structure and maps it to our internal structure. Not doing so intermixes conventions, and I'd like to avoid that.

We don't have to include the use of internal prefixes, but at least the core name. So, for example:

- `button-border-radius` versus `buttonRounded`
- `card-bg-color` versus `journeyCardBackgroundColor`
- `button-focus-ring-color` versus `buttonFocusBorderColor`
- etc.

apps/login-app/src/routes/(app)/+layout.svelte
```diff
@@ -1,12 +1,28 @@
...
 
+<script lang="ts">
+  import { browser } from '$app/environment';
+  import { applyThemeVars } from '$core/_effects/theme.effects';
+  import { styleStore } from '$core/style.store';
+
+  export let data: { backgroundImageUrl?: string };
+
+  $: if (browser) {
+    applyThemeVars(document.body, $styleStore?.theme);
```
cerebrl: Can you explain why you chose to apply the theme vars exclusively in the browser, rather than write them to a file on the server? I'm concerned about low power devices and a possible flash of style changes after initial render.

I'm also curious as to why the App utilizes a different approach than the Widget. It seems the Widget has its own hard-coded CSS file, but the app utilizes this more dynamic approach.

packages/login-widget/src/lib/custom.base.css
```diff
@@ -47,6 +48,62 @@
     Segoe UI Emoji,
     Segoe UI Symbol,
     Noto Color Emoji; /* 4 */
+
+  /* Default HSL channel pairs for every --tw-colors-* slot referenced in compiled
+     component classes. Values match the token hex values in tokens.cjs exactly so
+     the widget renders identically to before this declaration block was added.
+     Override these on .fr_widget-root at runtime (via applyThemeVars) to retheme
+     without touching compiled CSS. */
+  --tw-colors-background-dark-hs: 217.2, 32.6%;
```
cerebrl: Is this necessary? If customers use the Tailwind theme configuration, do they also have to update the values here too? I would think so, which seems sub-optimal at first glance.

cerebrl (changes requested):
Some comments after an initial scan.

vatsalparikh (changes requested):
Found a bug while testing the PR. Loading a page normally (without specifying a journey) picks up the tenant theme instead of the theme defined in the journey. The expected behaviour is that tenant theme is okay but overriding the journey theme does not work and journey theme is not picked up after override. Confirmed this bug with Gabriel.

apps/login-app/src/routes/(app)/+layout.server.ts
vatsalparikh: Would be good to add tests for fetchIdmTheme function

apps/login-app/src/routes/e2e/widget/modal/theme/+page.svelte
```diff
@@ -0,0 +1,63 @@
...
+
+ Copyright © 2025-2026 Ping Identity Corporation. All right reserved.
+
+ This software may be modified and distributed under the terms
+ of the MIT license. See the LICENSE file for details.
+
+ -->
+
+<script lang="ts">
+  import { onMount } from 'svelte';
```
vatsalparikh, line 11: It would be good to add inline tests as well here.

e2e/tests/widget/modal/widget-modal.theme.test.js
```diff
@@ -0,0 +1,53 @@
+/**
+ *
+ * Copyright © 2025-2026 Ping Identity Corporation. All right reserved.
+ *
+ * This software may be modified and distributed under the terms
+ * of the MIT license. See the LICENSE file for details.
+ *
+ **/
+
+import { expect, test } from '@playwright/test';
```
vatsalparikh, line 10: Would be good to add inline tests here too

vatsalparikh (commented):
Another review:

- Added file and line specific comments below
- PR wide comments:
	- Login widget is a public API. We should add documentation about the new theme property and its associated sub-properties wherever applicable.

apps/login-app/src/routes/(app)/+layout.svelte
SteinGabriel: Yeah, good point. I haven't considered that use case. The way I found to solve this was adding `themeToStyleTag` that generates an inline `<style>` block server-side. IDM theme vars are now in the initial HTML so there's no flash on first render. 
That works but doesn't sound ideal either, so I'm still working on the idea of having a static css file, instead of creating the css vars dynamically, to see how that works. I'll see if I can find a way to not duplicating the css files between widget and login-app.

apps/login-app/src/routes/(app)/+layout.server.ts
SteinGabriel: Renamed all `ThemeObject` fields to align with their CSS custom property targets (e.g. `buttonBorderRadius` → `--fr-button-border-radius`). `toIdmTheme` is now a proper translation layer that maps IDM's external field names to our internal naming conventions. Thanks!

apps/login-app/src/routes/(app)/+layout.server.ts
SteinGabriel: Yeah, good call. Extracted `fetchIdmTheme` and all supporting types/cache into `src/lib/server/idm-theme.effects.ts`. `+layout.server.ts` now only reads env vars and calls `fetchIdmTheme`

SteinGabriel:
> Found a bug while testing the PR. Loading a page normally (without specifying a journey) picks up the tenant theme instead of the theme defined in the journey. The expected behaviour is that tenant theme is okay but overriding the journey theme does not work and journey theme is not picked up after override. Confirmed this bug with Gabriel.

Fixed.  When no `?journey=` param is present,  it now falls back to `FR_AM_JOURNEY_LOGIN` env var, since there's no way to determine what is the tenant's default journey from "client" side. Make sure to set `FR_AM_JOURNEY_LOGIN=Login` in your `.env`. Thanks!

packages/login-widget/src/lib/custom.base.css
SteinGabriel: The hardcoded CSS var blocks are removed. The Tailwind plugin now auto-derives all `--tw-colors-*` defaults from `tokens.cjs` at build time. If customers change the Tailwind tokens, the CSS vars update automatically on next build. No manual sync needed anymore. 
Thanks for pointing this out!

apps/login-app/src/routes/(app)/+layout.svelte
SteinGabriel: IDM theme vars are now applied server-side as an inline style attribute on the root wrapper div, so they're in the initial HTML with no flash. The `if (browser)` DOM approach has been removed. `{@html}` has also been removed in favor of a normal style attribute binding on a wrapper div, and the `--tw-colors-*` default values are now auto-derived from `tokens.cjs` by the Tailwind plugin at build time.

apps/login-app/src/routes/(app)/+layout.server.ts
SteinGabriel: For sure! I added `idm-theme.effects.test.ts` with tests covering the full resolution logic, cache fallback paths, and graceful degradation for network errors, bad status codes, and invalid json. Thanks!

apps/login-app/src/routes/e2e/widget/modal/theme/+page.svelte
SteinGabriel: Added, thanks!

e2e/tests/widget/modal/widget-modal.theme.test.js
SteinGabriel: Added, thanks!

cerebrl (approved):
This looks good to me.

.env.example
```diff
@@ -5,6 +5,7 @@ FR_REALM_PATH=<<<(string) Realm path of AM>>>
 FR_OAUTH_PUBLIC_CLIENT=<<<(string) Your Public OAuth client name/ID>>>
 # FR_OAUTH_SCOPE=openid profile me.read  # optional — AM uses client defaults if omitted
 FR_AM_WELLKNOWN_URL=<<<(string) Your OIDC Discovery Wellknown Endpoint>>>
+# FR_AM_JOURNEY_LOGIN=Login  # optional — default login journey name; used for IDM theme resolution when no ?journey= param is present
```
vatsalparikh, line 8: I do understand the idea behind specifying an .env variable.

However, I am thinking something else. There is always a default journey that is specified in AM. Is it possible to retrieve that default journey somehow? Is that a separate AM request? Would be great to explore that side of things, because this will help get the theme for that default journey without messing with .env variables.

In fact, for an application that's deployed in cloud or somewhere else this creates a burden on specifying the journey name in .env.

My point is, the default journey should be known to login app so it can automatically apply the themes as soon as the application loads.

This may be significant scope creep I guess, but worth exploring if it's easy to implement or maybe in another ticket

themes/default/token-vars.cjs
```diff
@@ -0,0 +1,70 @@
...
+  ['colors.secondary.light', 'secondary-light'],
+  ['colors.success.dark', 'success-dark'],
+  ['colors.success.light', 'success-light'],
+  ['colors.tertiary.dark', 'tertiary-dark'],
+  ['colors.tertiary.light', 'tertiary-light'],
+  ['colors.warning.dark', 'warning-dark'],
+  ['colors.warning.light', 'warning-light'],
+];
+
+function hexToHslChannels(hex) {
```
vatsalparikh, line 40: This function `hexToHslChannels ` is a duplicate of the function defined in theme.utilities file. Is this on purpose or a mistake?

core/_utilities/theme.utilities.ts
```diff
@@ -0,0 +1,208 @@
...
+    cssVar('--fr-page-bg-color', theme.backgroundColor);
+  }
+  if (theme.linkColor) {
+    hslPair('--tw-colors-link-dark-hs', '--tw-colors-link-dark-l', theme.linkColor);
+  }
+  if (theme.linkActiveColor) {
+    hslPair('--tw-colors-link-light-hs', '--tw-colors-link-light-l', theme.linkActiveColor);
+  }
+  if (theme.fontFamily) {
+    cssVar('--fr-font-family', theme.fontFamily);
```
vatsalparikh, line 160: Font Family is not working. I changed the font family to Stalemate and it was not picked up as shown in the theme, and not visible when I spin up the app locally. Screenshots shared.

Image

Image

Here is a potential fix:
The font family pipeline is correctly written end-to-end (fetch → parse → sanitize → CSS var), but nothing actually calls setProperty on the DOM. Three wiring steps are missing:

[+layout.server.ts](vscode-webview://1gicedcv2vvr6dvvsot8sb4c0r8e3tfcd99f3gc7mg77d23dt19t/apps/login-app/src/routes/(app)/+layout.server.ts) — no fetchIdmTheme() call, so fontFamily is never fetched from AM server-side.
[+layout.svelte](vscode-webview://1gicedcv2vvr6dvvsot8sb4c0r8e3tfcd99f3gc7mg77d23dt19t/apps/login-app/src/routes/(app)/+layout.svelte) — no script reading server data and applying CSS vars to the DOM.
[index.svelte](vscode-webview://1gicedcv2vvr6dvvsot8sb4c0r8e3tfcd99f3gc7mg77d23dt19t/packages/login-widget/src/lib/index.svelte) — no reactive applyThemeVars call on the widget root element.
Because none of these fire, var(--fr-font-family, 'Open Sans') always falls back to Open Sans — regardless of what's set in AM.

There's also a test gap: the E2E fixtures don't assert --fr-font-family is ever set on .fr_widget-root, so this would silently regress.

vatsalparikh (changes requested):
**Only major request is font family not working**

Some other non-blocking comments

I noticed we are not applying favicons. Confirmed with Gabriel that it is not in scope for this PR. We should include it in the next phase of theming implementation

core/_utilities/theme.utilities.ts
SteinGabriel: That should be fixed for both login-app and login-widget. Thank you!

themes/default/token-vars.cjs
SteinGabriel: Yeah, this duplication is intentional. `token-vars.cjs` is a CommonJS file used as a Tailwind plugin and can't cleanly import from `theme.utilities.ts`. The two files live in separate module systems. I guess it'd be technically more doable to have a shared .`cjs` file being imported into both `.ts` and `.cjs` files, but I don't think importing a function into `core/` from `theme/` is a pattern we'd like to introduce.

.env.example
SteinGabriel: Retrieving the default tenant's journey automatically would require an authenticated call to the AM realm config endpoint. It's doable but adds auth complexity and a new failure mode to the SSR path. I guess we could look into this in a future version.

core/_utilities/theme.utilities.ts
vatsalparikh: I did try downloading a font in macos but was not picked up. Maybe the naming or something else in my local system.

Either way, theming is a multi-phase and multi-sprint effort, so we're good here and add / modify things later.

vatsalparikh (approved):
Looks great!

Theming is a multi-phase and multi-sprint effort, so we're good here and we can add / modify things later.

PR #536, opened 2026-06-22
fix(deps): resolve duplicate postcss-selector-parser override

SteinGabriel:
The theming PR added a second `postcss-selector-parser` override entry to `package.json` instead of updating the existing one. This produced a duplicate mapping key in `pnpm-lock.yaml`, breaking `pnpm install --frozen-lockfile` in CI.

The original `>=7.0.0` override was also ineffective against CVE-2026-9358, which affects`7.1.1`. The fix consolidates to a single `>=7.1.2` override, which resolves to `7.1.4` (the current patched release).

  ### Changes
  - `package.json`: removed duplicate `postcss-selector-parser` key, set to `>=7.1.2`
  - `pnpm-lock.yaml`: deduplicated override key, resolves `postcss-selector-parser@7.1.4`

  ### Fixes
  - `ERR_PNPM_BROKEN_LOCKFILE: duplicated mapping key` in CI
  - CVE-2026-9358 (medium, CVSS 4.3) via transitive `tailwindcss → postcss-selector-parser`

  ### Tested
  - `pnpm install --frozen-lockfile` passes locally
  - 244 unit tests pass
  - 54 E2E tests pass (2 skipped — pre-existing, unrelated)
  - Manually verified install and build complete without errors

### DIFF
--- a/package.json
+++ b/package.json
@@ -101,10 +101,9 @@
       "picomatch@2": "2.3.2",
       "picomatch@4": "4.0.4",
       "uuid": "^14.0.0",
-      "postcss-selector-parser": ">=7.0.0",
+      "postcss-selector-parser": ">=7.1.2",
       "ws": ">=8.21.0",
-      "yaml": ">=2.8.3",
-      "postcss-selector-parser": ">=6.1.4"
+      "yaml": ">=2.8.3"
     }
   }
 }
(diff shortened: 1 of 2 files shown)


vatsalparikh (approved):
Thanks!

The latest version for postcss selector parser is 7.1.4 so hopefully, we should have mend passing correctly now.

PR #542, opened 2026-06-29
SDKS-5186: Fix Alert Info Visual Regression From IDM Theming

SteinGabriel:
## Summary

  https://pingidentity.atlassian.net/browse/SDKS-5186

  The IDM theming feature introduced `buildTokenVars`, which injects `--tw-colors-*` CSS custom properties globally. This activated three previously-inert `calc()` lightness expressions in the Alert:Info component, producing an invalid green background in light mode and a black background in dark mode. This PR replaces those additive offsets with multiplicative `calc(var(--l) * factor)` expressions that correctly replicate the `color` library's relative `lighten`/`darken` math for any token value.

  ## Changes

  ### `themes/default`

  - `primitives.cjs` — added inline `relativeLightness(lightnessVar, op, ratio)` helper; replaced additive `+ 90%` / `-40%` offsets at `.alert-info` bg (light mode), `.alert-info_dark` bg, and `.alert-info_dark svg` color with
  multiplicative `calc(var(...) * factor)`  - `.changeset/fix-alert-info-calc-lightness.md` — patch changeset documenting the bug fix

  ### `.github/workflows`

  - `ci.yml` — added `untraced: 'themes/default/**'` to Chromatic action; prevents TurboSnap from reusing stale snapshots when only theme files change (false-negative: 0 affected stories traced → fix invisible in CI)

  ## Tests

  - 207/207 Storybook interaction tests pass after `pnpm build:widget`
  - `pnpm check:lint` clean

  ## How to test

  ### 1. Build and open Storybook

  ```sh
  pnpm build:widget
  pnpm storybook

  Navigate to Alert → Info story. In light mode the background should be light blue (not green). In dark mode it should be dark navy (not black).

  2. Verify all other alert variants unchanged

  Check Error, Success, and Warning stories in both light and dark — no visual change expected.

  Verify Chromatic (CI)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,7 @@ jobs:
       - uses: chromaui/action@latest
         with:
           onlyChanged: true
+          externals: 'themes/default/**'
           exitOnceUploaded: true
           traceChanged: true
           diagnosticsFile: chromatic-diagnostics.json
--- a/themes/default/primitives.cjs
+++ b/themes/default/primitives.cjs
@@ -8,6 +8,8 @@
  **/
 
 const colorLib = require('color');
+const { relativeLightness } = require('./lightness.cjs');
+
 module.exports = function (theme) {
   return {
     /**
@@ -74,8 +76,11 @@ module.exports = function (theme) {
     },
     '.alert-error_dark': {
       '--bg-color': 'hsl(var(--tw-colors-error-dark-hs),var(--tw-colors-error-dark-l))',
-      '--border-color':
-        'hsl(var(--tw-colors-error-dark-hs), calc(var(--tw-colors-error-dark-l) - 20%))',
+      '--border-color': `hsl(var(--tw-colors-error-dark-hs), ${relativeLightness(
+        '--tw-colors-error-dark-l',
+        'darken',
+        0.2,
+      )})`,
       backgroundColor: `var(--bg-color, ${theme('colors.error.dark')})`,
       borderColor: `var(--border-color, ${colorLib(theme('colors.error.dark'))
         .darken(0.2)
@@ -88,8 +93,11 @@ module.exports = function (theme) {
       },
     },
     '.alert-info': {
-      '--bg-color':
-        'hsl(var(--tw-colors-primary-light-hs), calc(var(--tw-colors-primary-light-l) + 90%))',
+      '--bg-color': `hsl(var(--tw-colors-primary-light-hs), ${relativeLightness(
+        '--tw-colors-primary-light-l',
+        'lighten',
+        0.9,
+      )})`,
       '--border-color': 'hsl(var(--tw-colors-primary-light-hs),var(--tw-colors-primary-light-l))',
       backgroundColor: `var(--bg-color, ${colorLib(theme('colors.primary.light'))
         .lighten(0.9)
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -158,6 +158,7 @@ jobs:
       - uses: chromaui/action@latest
         with:
           onlyChanged: true
+          untraced: 'themes/default/**'
```
vatsalparikh: What is the purpose of adding untraced here?

I looked at the documentation for untraced here https://www.chromatic.com/docs/configure/ and it says that `Disregard these files and their dependencies when tracing dependent stories`. Does that mean we are avoid snapshots for dependents of all files under themes/default/** folder?

It looks like you were trying to clear cache / stale snapshots from your PR description but in reality I think those snapshots are completely skipped. Might be good to dive into this further I think.

themes/default/primitives.cjs
```diff
@@ -8,6 +8,13 @@
  **/
 
 const colorLib = require('color');
+
+function relativeLightness(lightnessVar, op, ratio) {
```
vatsalparikh: There is enough logic in here to add unit tests for this `relativeLightness`

themes/default/primitives.cjs
vatsalparikh, line 1: Should we look into `.alert-error_dark`, `.alert-success_dark`, and `.alert-warning_dark` classes as well, whether we need to change lightness calculation to multiplicative instead of additive here?

vatsalparikh (changes requested):
Some changes requested and some questions.

.github/workflows/ci.yml
SteinGabriel: Good catch! It was supposed to be `externals` instead of `untraced`. The idea is to [force a full re-test](https://www.chromatic.com/docs/turbosnap/setup/#specify-external-files-to-trigger-a-full-re-test-when-they-change) of the files in the `themes/default/` directory, which lives outside the webpack's dep graph. This was causing the Chromatic check to be cached for those `.cjs` files. It looks like it's picking up the latest changes correctly now. Thank you!

themes/default/primitives.cjs
SteinGabriel: Fair point. I was avoiding a new file just for one function, but having test coverage is worth it for sure. 
Added. Thanks!

themes/default/primitives.cjs
```diff
@@ -103,8 +113,11 @@ module.exports = function (theme) {
...
     '.alert-info_dark': {
-      '--bg-color':
-        'hsl(var(--tw-colors-primary-light-hs), calc(var(--tw-colors-primary-light-l) - 40%))',
+      '--bg-color': `hsl(var(--tw-colors-primary-light-hs), ${relativeLightness(
+        '--tw-colors-primary-light-l',
+        'darken',
+        0.4,
+      )})`,
       '--border-color':
         'hsl(var(--tw-colors-primary-dark-hs), calc(var(--tw-colors-primary-dark-l) - 20%))',
```
vatsalparikh: Should we also change `--tw-colors-primary-dark-l` from additive to multiplicative?

themes/default/primitives.cjs
SteinGabriel: Right, I missed them since those cases had no issue calculating their color values. 
Updated. Thanks for pointing it out.

themes/default/primitives.cjs
SteinGabriel: Updated. Thanks!

vatsalparikh (approved):
Looks great!

Reviewed code and tested changes in Storybook!

PR #552, opened 2026-07-17
refactor(oidc-and-protect): refactor login framework to use the new oidc and protect clients from Ping Orch SDKs

vatsalparikh:
## JIRA Ticket
https://pingidentity.atlassian.net/browse/SDKS-5216

## Technical decisions
- Using normal oidc client config instead of unified config because it’s easier, avoids importing sdk utils package
- Defaults for OIDC
	- scope defaults to 'openid' in the Zod schema
	- redirectUri, clientId, serverConfig.wellknown are all required, no defaults provided
- Cleaned up login app's register function. We no longer need oauth or user in login app, so that logic is completely removed from login app

## Tests performed
- Sample app login widget
	- [x] 	- login flow, todo loading
	- [x] 	- Verified local storage and cookies to check both journey client and oidc client flows
	- [x] 	- Verified logout needs both, journey client terminate and oidc client user logout. Journey client terminate removes cookies, oidc user logout removes local storage data
	- [x] 	- Protect tests pass

- Login Framework
	- [x] 	- Login app works as expected (however it only uses journey client so no impact whatsoever)
	- [x] 	- e2e widget modal login and logout works as expected
	- [x] 	- e2e widget inline login works, logout fails (it failed in main branch too, this is an existing issue, unrelated to journey client terminate. The AM session cookie isn't sent when terminate() makes a cross-origin request from https://localhost:8443 to openam-sdks.forgeblocks.com. The CORS config is fine and third-party cookies are allowed in the browser, but the AM session cookie itself has SameSite restrictions that prevent it being included in the cross-origin POST.)
	- [x] 	- Added back (unskipped) protect tests, both initialization and evaluation tests pass

## How to test
Pull https://github.com/ForgeRock/sdk-sample-apps/tree/migrate-login-widget-2.0 branch locally and test journey, oidc, and protect flows.
Here is the PR that accommodates all the updates necessary in the sample app https://github.com/ForgeRock/sdk-sample-apps/pull/123

## Recordings
The recordings below show how both journey client terminate and oidc user logout are necessary in user.logout function in login widget. They both perform two different operations. The recordings also show how tokens are correctly stored in cookie and local storage respectively, verifying that login, tokens, logout all work as expected.
1. Removed OIDC user logout: https://github.com/user-attachments/assets/881192d3-4bba-41e1-9e79-da20c768908a
2. Removed Journey Terminate: https://github.com/user-attachments/assets/6b60b73b-ec49-4e17-92aa-89e5515e2026
3. Happy Path tokens and logout work as expected: https://github.com/user-attachments/assets/e4da1712-bc30-46be-bc04-2719dd812cfc


core/oauth/oauth.store.ts
```diff
@@ -57,31 +88,62 @@ function getTroubleshootingMessage(message: string) {
...
+   * Returns a cached, ready-to-use OIDC client promise.
+   * Exposed on the returned store so `user.store` and `widget.api` can inject it
+   * without reaching back into module scope.
    */
-  async function get(getOptions?: GetTokensOptions) {
-    /**
-     * Create an options object with getOptions overriding anything from initOptions
-     * TODO: Does this object merge need to be more granular?
-     */
+  async function getOidcClient(): Promise<OidcClientReady> {
```
cerebrl: I'm wondering if we could improve some of this composition by not mixing Promises with Svelte Stores. Have you tried wrapping the SDK OIDC Client initialization in a Svelte Store, rather than wrapping it in an `async` function? Then, we can compose this larger, framework store with this smaller, nested SDK store.

core/oauth/oauth.store.ts
```diff
@@ -91,30 +154,43 @@ export function initialize(initOptions?: GetTokensOptions) {
     });
 
     try {
-      tokens = await TokenManager.getTokens(options);
-    } catch (err: unknown) {
-      if (err instanceof Error) {
+      const oidcClient = await getOidcClient();
+      const tokens = await oidcClient.token.get(options);
```
cerebrl: This is what's feeling a bit odd for me. My intuition tells me that this function should already have an SDK OIDC Client referenced, removing the need to create one asynchronously like this. I'm wondering if converting this Promise to a Svelte Store would provide something akin to `sdkOidcStore` and just calling methods off of the store, rather than awaiting Promises and then calling `token.get` off of the returned Promise value.

core/oauth/oauth.store.ts
```diff
@@ -128,7 +204,8 @@ export function initialize(initOptions?: GetTokensOptions) {
   }
 
   return {
-    get,
+    getTokens,
```
cerebrl: Did you change this to `getTokens` because you needed to add `getOidcClient`? You had to "get" methods?

core/protect/protect.store.ts
cerebrl, line 1: The term "store" in this context is supposed to mean a Svelte Store, kind of like a "store" in the SDK is a Redux Store. They are both implementations of the Observable pattern. Can we convert this into a Svelte Store like the other `store.ts` files?

core/user/user.store.ts
```diff
@@ -70,7 +60,21 @@ export function initialize(initOptions?: ConfigOptions) {
     });
 
     try {
-      const user = await UserManager.getCurrentUser(options);
+      const oidcClient = await getOidcClient();
```
cerebrl: Yeah, I think we should try to avoid this pattern. I this converting this Promise into a store that both the framework level OAuth and this User Store can rely on would be better.

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -69,38 +72,14 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
-    }
-
-    /**
-     * Initialize all the stores.
-     */
+    // initialize() creates a fresh, isolated store instance per call.
+    // getOidcClient is injected into the user store so it shares the same
+    // OIDC client instance without reaching into module-level state.
+    oauthStore = initializeOauth(options?.oidcClient);
+    userStore = initializeUser(oauthStore.getOidcClient);
```
cerebrl: Just reiterating here: I'd like to avoid this pattern; mixing Promises with Svelte Stores. We can either create and export a store from a module that represents a wrapped SDK client, or export an `initialize` function that creates a store in a closure, and pass that store into `initializeOauth` and `initializeUser`.

cerebrl (changes requested):
I'd like to stick with the pattern of using Svelte Stores for managing state and it's dependencies. Svelte refers to these as "custom stores" or "derived stores". You can see examples of them in this repo. Here's one that's fairly simple: https://github.com/ForgeRock/forgerock-web-login-framework/blob/main/packages/login-widget/src/lib/_utilities/component.utilities.ts.

core/protect/protect.store.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @returns {Promise<void | { error: string }>}
+ */
+export function start(config: ProtectConfig | SignalsInitializationOptions) {
+  /**
+   * Accepts either a `ProtectConfig` (has `envId` — validated by Zod) or a
+   * `SignalsInitializationOptions` (from `callback.getConfig()` — passed through directly).
+   * The underlying `protect()` factory handles both shapes natively.
+   */
+  const safeConfig: SdkProtectConfig | SignalsInitializationOptions =
+    'envId' in config ? protectConfigSchema.parse(config) : config;
```
SteinGabriel: Determining whether `protectConfigSchema.parse(config)` is called whenever `envId` in `config` sounds a bit fragile. What happens when `envId` is present, but the rest of the config object is malformed? For example, if another key in `config` is not recognized by `protectConfigSchema.strict()`. Since it's not wrapped in a `try/catch`, it would probably result in an uncaught error. Would it be possible to call `.parse(config)` based on the caller of this function instead? That would be more deterministic.

core/oauth/oauth.store.ts
vatsalparikh: Yes, it would be confusing otherwise with a normal get, changed it to getTokens to improve readability

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -279,40 +231,38 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
+        const oidcClient = await oauthStore.getOidcClient();
+        await oidcClient.user.logout();
+      } catch (err: unknown) {
+        // Warn instead of throw; throwing to the caller would be misleading since there's nothing to recover from.
+        console.warn('OIDC logout failed:', err instanceof Error ? err.message : err);
       }
-      // Return undefined as there's no response information to share
-      return;
+      // Regardless of errors, reset all stores and restart journey
+      resetAndRestartStores();
```
SteinGabriel: I'm not sure about silently failing with only `console.warn` here. What happens when the stores are reset and restored, but the server-side session was not actually terminated? 
Also, this function can still throw an `Error` via `logErrorAndThrow()` on line 241, which is not documented in the function's JSDoc block btw.

core/journey/callbacks/ping-protect-initialize/ping-protect-initialize.svelte
```diff
@@ -29,15 +29,12 @@
...
+        callback.setClientError(result.error);
+      } else {
         loaded = true;
-      } catch (error) {
-        if (error instanceof Error) {
-          callback.setClientError(error.message);
-        } else {
-          callback.setClientError('An error occurred while initializing PingProtect');
-        }
+        console.log('Protect initialized by callback for data collection');
```
SteinGabriel: Is this `console.log` intentional?

core/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
```diff
@@ -8,49 +8,47 @@
...
-          callback.setClientError(error.message);
-        } else {
-          callback.setClientError('An error occurred while initializing PingProtect');
-        }
+      const result = await getData();
+      if (typeof result === 'object') {
+        callback.setClientError(result.error);
+      } else {
+        callback.setData(result);
+        console.log('Data set on Protect evaluation callback');
```
SteinGabriel: Is this `console.log` intentional?

core/journey/callbacks/ping-protect-initialize/ping-protect-initialize.svelte
vatsalparikh: Yes, it is used in the test and is an indicator that protect was initialized successfully! I think it's fine to log it in the console but I will check if we can avoid this.

core/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
vatsalparikh: Yes, it is used in the test and is an indicator that protect was evaluated successfully! I think it's fine to log it in the console but I will check if we can avoid this.

ryanbas21: This would log in someone's app code then, right?

core/journey/callbacks/ping-protect-initialize/ping-protect-initialize.svelte
ryanbas21: This would also log in a live application right?

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: > What happens when the stores are reset and restored, but the server-side session was not actually terminated?

I think the flow for this would be, if logout fails and we reset and restart stores, as soon as the user clicks login, they will be automatically logged in because their server side session is still valid. On the client side, security wise it's good to reset and restart all stores. We are informing the user (by throwing) that there was an error logging out so the user can expect to be automatically logged on even after reset if they know that logout failed.

Some other things:
- I realized reset and restart was only happening in the catch block. I've restored code to match main where reset and restart happens in both try and catch block.
- I have removed console warn and am throwing just like we did before in the main branch.
- Added jsdoc that we are throwing error in this function

Thank you very much for pointing this out! Lots of issues there that needed to be addressed 🙏

vatsalparikh:
> I'd like to stick with the pattern of using Svelte Stores for managing state and it's dependencies. Svelte refers to these as "custom stores" or "derived stores". You can see examples of them in this repo. Here's one that's fairly simple: [`main`/packages/login-widget/src/lib/_utilities/component.utilities.ts](https://github.com/ForgeRock/forgerock-web-login-framework/blob/main/packages/login-widget/src/lib/_utilities/component.utilities.ts).

Really appreciate your feedback here. Took me several iterations to fully understand and implement a declarative, derived, observer based svelte store. Thank you so much for taking the time to review it deeply and providing all the suggestions. It was a challenging and very rewarding experience!

I have created an oidc store for oidc client initialization. widget.api file composes it and passes it to both oauth and user stores. They consume the client. The code is ready for another review now. I believe some minor tweaks can be made but the code is structurally sound now in the sense that I've avoided promises, race conditions, and other issues with the help of declarative, observable pattern based code!

The PR is ready for review again, @cerebrl !

core/oauth/oauth.store.ts
vatsalparikh: Thank you 🙏 

Created oidc.store.ts file and a writable svelte store for oidc client, which is then used by widget.api and passed to both oauth.store and user.store.

core/oauth/oauth.store.ts
vatsalparikh: Thank you 🙏 

Passing the oidc client store from widget.api, so this is already available. We still check whether it is available or not to correctly populate oauth store.

core/oauth/oauth.store.ts
vatsalparikh: Unnecessary now, back to using the normal `get` in oauth.store

core/protect/protect.store.ts
vatsalparikh: Thank you 🙏 

Created a writable protectClientStore

core/user/user.store.ts
vatsalparikh: Thank you 🙏 

Using svelte's get to retrievethe oidc client store.

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: Thank you 🙏

Reactively making sure that oidc client store has the oidc client initialized by using derived from svelte.

core/protect/protect.store.ts
vatsalparikh: Thank you 🙏

Moved the check to zod validation. The only drawback is because we accept both types of protect configs, any incorrect env id like a misspelled variable is still accepted because protect config accepts record<unknown, string>. If env id is present, it will always validate.

core/journey/callbacks/ping-protect-initialize/ping-protect-initialize.svelte
vatsalparikh: Removed, unnecessary now

core/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
vatsalparikh: Removed, unnecessary now

vatsalparikh: Removed, unnecessary now

core/journey/callbacks/ping-protect-initialize/ping-protect-initialize.svelte
vatsalparikh: Removed, unnecessary now

packages/login-widget/vite.config.ts
```diff
@@ -79,9 +79,9 @@ export default defineConfig({
     rollupOptions: {
       // Externalize runtime dependencies — they're provided by the consumer
       external: [
-        '@forgerock/javascript-sdk',
         '@forgerock/journey-client',
-        '@forgerock/ping-protect',
+        '@forgerock/oidc-client',
+        '@forgerock/protect',
```
ryanbas21: Let's move these out of here, move the package installs into the root package.json

ryanbas21: Leave the rest of the items in the array, as I remove those in the other PR i have open

core/protect/protect.store.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+/**
+ * @function getData - Returns device data from the shared Protect client.
+ * @returns {Promise<string | { error: string }>}
+ */
+function getData() {
+  const protectClient = get(protectClientStore);
+  if (!protectClient) {
+    return Promise.resolve({ error: 'PingOne Signals SDK is not initialized' });
```
ancheetah: Should these error messages say `Protect client not initialized` instead? I say this because all of the Protect methods can already return a `Signals SDK not initialized` error so this would help distinguish what is coming from the login widget vs protect sdk. FWIW I don't believe it's possible to not get back a protect client. It's a synchronous factory function that returns an API and doesn't throw.

core/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
```diff
@@ -8,49 +8,46 @@
...
-      try {
-        await PIProtect.getData();
-      } catch (error) {
-        if (error instanceof Error) {
-          callback.setClientError(error.message);
-        } else {
-          callback.setClientError('An error occurred while initializing PingProtect');
-        }
+      const result = await protectStore.getData();
+      if (typeof result === 'object') {
```
ancheetah: Checking for `error` in result makes this a little more explicit and clear but nbd. I think technically `null` passes through here too but that shouldn't happen.

core/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
```diff
@@ -8,49 +8,46 @@
...
-    pauseBehavioralData = pingProtect?.behavioralDataCollection ?? false;
+    isBehavioralDataPaused = pingProtect?.behavioralDataCollection ?? false;
     if (typeof window !== 'undefined') {
-      if (pauseBehavioralData === true) {
-        PIProtect.pauseBehavioralData();
+      if (isBehavioralDataPaused === true) {
+        protectStore.pauseBehavioralData();
       } else {
-        PIProtect.resumeBehavioralData();
+        protectStore.resumeBehavioralData();
```
ancheetah: Should check for `error` in the result here as well.

ancheetah (commented):
Some comments on Protect. Is there existing e2e for Protect as well?

core/protect/protect.store.ts
SteinGabriel: Thanks for updating this. Since `protectConfigSchema` is `.strict()` and `signalsInitializationOptionsSchema` is `z.record<z.string(), z.string()`, any variable included in `config` that is not part of the protect config schema and its vale is not a string (e.g `{ envId: '1234', consoleLogEnabled: true }`),  will throw a `ZodError`. I'm not seeing where the protect config schema accepts `Record<unknown, string>`, so please let me know if I'm missing something. If what I said is correct, throwing a `ZodError` from `protectConfigSchema` would result in an uncaught error where `protectStore.start(config)` is called (`ping-protect-initialize.svelte:32`). 
That being said, we could either remove `.strict()` so only the fields the widget cares about are validated, wrap the callback's await `protectStore.start(config)` in a try/catch or change `signalsInitializationOptionsSchema` validation, if that's the right call.

packages/login-widget/src/lib/types.ts
```diff
@@ -34,28 +32,24 @@ export type ConfigurationApi = ReturnType<typeof _api.configuration>;
...
 export type JourneyStoreValue = JourneyStoreEventValue;
 export type OAuthTokenStoreValue = OAuthTokenStoreEventValue;
 export type UserStoreValue = UserStoreEventValue;
 
-// SDK Configuration Options Type
-export type ConfigOptions = SdkConfigOptions;
-
 // SDK OAuth Tokens Type
-export type OAuth2Tokens = SdkOAuth2Tokens;
+export type OauthTokens = SdkOauthTokens;
```
cerebrl, line 50: Should we use OIDC, rather than OAuth to align with the rest of our naming conventions, like `OidcTokens`, rather than `OauthTokens`?

vatsalparikh: I think we should keep using OauthTokens because SdkOauthTokens is an alias for the SDK type OauthTokens.

`import type { OauthTokens as SdkOauthTokens } from '@forgerock/oidc-client/types';`

So this is identical to SDK.

packages/login-widget/vite.config.ts
vatsalparikh: done

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -68,91 +69,50 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
+   */
+  async function configure(options?: WidgetConfigOptions): Promise<void> {
+    const wellknown = options?.wellknown;
+    const journeyClientConfig = wellknown ? { serverConfig: { wellknown } } : undefined;
+
+    if (options?.oidcClient) {
+      // The OIDC client cannot be constructed without a well-known URL. Fail fast rather
+      // than silently skip construction and leave token/user/logout APIs unusable.
+      if (!wellknown) {
+        throw new Error('`wellknown` is required when `oidcClient` is configured.');
```
cerebrl: Is this still true? Doesn't the Journey Client also require a wellknown? As it is now, I would believe that `wellknown` is a required property regardless, yeah?

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -68,91 +69,50 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
-        /**
-         * Initialize the stores and ensure both variables point to the same reference.
-         * Variables with _ are the reactive version of the original variable from above.
-         */
-        journeyStore = initializeJourney(setOptions?.journeyClient, {
-          ...(setOptions?.captcha && { captcha: captchaConfigSchema.parse(setOptions.captcha) }),
-        });
-        oauthStore = initializeOauth(setOptions?.forgerock);
-        userStore = initializeUser(setOptions?.forgerock);
+    const [oidcClient] = await Promise.all([oidcPromise, journeyPromise]);
```
cerebrl: Do we not care about the result of the `journeyPromise`? Can it also fail? If it does, the Login Widget can't perform its primary duty.

vatsalparikh: Oidc client never rejects, always resolves, so we need to destructure and handle it after Promise.all call.

Journey client can throw so the promise can be rejected. When the journey client resolves, its value is already cached in the variable, so we don't have to destructure it because we aren't doing anything with that value after the promise.all call.

https://github.com/ForgeRock/forgerock-web-login-framework/blob/6692ca48f9015483f91df3d426130f75ae7d97c4/core/journey/journey.store.ts#L118

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -68,91 +69,50 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
-        ...options?.forgerock,
-        // Force 'legacy' to remove confusion
-        ...{ support: 'legacy' },
+  /**
+   * @function configure - Configures the widget and constructs its clients.
+   * @param {WidgetConfigOptions} options - The configuration options for the widget
+   * @returns {Promise<void>} Resolves when both clients are constructed
+   * @throws {Error} If config is invalid (Zod) or either client fails to construct
+   */
+  async function configure(options?: WidgetConfigOptions): Promise<void> {
```
cerebrl: I think we have too much optional. At this point, `options` needs to be required. Without it, the Widget can't do anything, yeah? Let's make `options` required as well as the `wellknown` property.

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -68,91 +69,50 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
+    const wellknown = options?.wellknown;
+    const journeyClientConfig = wellknown ? { serverConfig: { wellknown } } : undefined;
```
cerebrl: Rather than this, let's just enforce the property `wellknown` as required. If it's not provided, we `throw`. 
The point of the Widget is rendering Journeys, if we a `wellknown` is not provided, we can't create a Journey Client.

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -68,91 +69,50 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
+
+    if (options?.oidcClient) {
```
cerebrl: The Widget can operate without OIDC, that is fine. Customers can use the Widget just for rendering Journeys. But the opposite is not true. You should NOT use the Widget just for OIDC. If you not rendering Journeys, just use the SDK directly. The Widget provides not value for OIDC only operation.

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -68,91 +69,50 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
-            // Let user provided config override defaults
-            ...setOptions?.forgerock,
-            // Force 'legacy' to remove confusion
-            ...{ support: 'legacy' },
-          });
-        }
+    // oidcClientStore.getClient() resolves with the client or a GenericError shape (never rejects);
+    // getJourneyClient() can reject — Promise.all propagates that as a rejected configure().
+    const oidcPromise = oidcClientStore ? oidcClientStore.getClient() : Promise.resolve(null);
+    const journeyPromise = journeyClientConfig ? getJourneyClient() : Promise.resolve(null);
```
cerebrl: This is essentially an error case that should be thrown, not handled with a default `Promise.resolve(null)`.

packages/login-widget/src/lib/widget.api.ts
```diff
@@ -68,91 +69,50 @@ export function widgetApiFactory(componentApi: ReturnType<typeof _componentApi>)
...
-            },
-            // Let user provided config override defaults
-            ...setOptions?.forgerock,
-            // Force 'legacy' to remove confusion
-            ...{ support: 'legacy' },
-          });
-        }
+    // oidcClientStore.getClient() resolves with the client or a GenericError shape (never rejects);
+    // getJourneyClient() can reject — Promise.all propagates that as a rejected configure().
+    const oidcPromise = oidcClientStore ? oidcClientStore.getClient() : Promise.resolve(null);
```
cerebrl: This is internal implementation detail, so I'm not too concerned with it for this PR. But, I do want to mention the inconsistency between this line and the one below does have a code smell to me. Being that both of these are Clients backed by stores, I feel they should have the same implementation design, i.e.:

```
oidcClientStore.getClient()
journeyClientStore.getClient()
```

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: If we make wellknown required, that will make things harder. For example, protect doesn't need wellknown. Someone only using protect might have to provide a wellknown.

I did think about making wellknown required, but my concern is, do we anticipate any package / client to be added later to login widget which doesn't require wellknown. Is it the right thing to do, to require wellknown?

vatsalparikh: I think your next comments answer my question, the fact that login widget is primarily for journeys and then optionally can be used with OIDC.

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: At the moment, journey client config is still a variable. When we move journey client into a store, we can correctly rename `getJourneyClient` to `getClient` because then we'll have an actual store.

If we do that now, it will become confusing because we can directly call getJourneyClient as a function rather than a property of a store.

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: Done. This is unnecessary now that we always attempt to initialize journey client.

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: Done. options is required, wellknown is required, and we always attempt to initialize journey client now.

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: done

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: done

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: Removed that piece of code, it's irrelevant now

packages/login-widget/src/lib/widget.api.ts
vatsalparikh: done, wellknown is required

core/protect/protect.store.ts
vatsalparikh: Sorry, I meant Record<string, string>, not unknown, and my response there was going nowhere.

I was thinking deeply about this, and your comment made me wonder which configs we are accepting through public API. I realized that in interfaces.ts file I am only accepting ProtectConfig. I have changed it to accept both ProtectConfig and SignalsInitializationOptions now!

Thanks for pointing out the strict solution. Routing to the parse function (using env id or something else) and strict are two different things. Removing strict is necessary because SDK's schema has limited number of fields, so our validation can return false negative even when the fields are valid. So I've removed strict. Instead, using looseObject so that any fields that are present in the schema are validated, whereas missing fields from schema are passed through. I’ve added comments in the code for future reference.

To your point about using envId, which was originally used to differentiate between ProtectConfig and SignalsInitializationOptions, z.union is doing the differentiation already, so that is solved!

Thanks for the detailed feedback!
All the issues with env id differentiation, strict, and thrown errors should be addressed now!

vatsalparikh:
> Some comments on Protect. Is there existing e2e for Protect as well?

Yes, there are some e2e protect tests here: https://github.com/ForgeRock/forgerock-web-login-framework/blob/SDKS-5216/core/protect/protect.store.test.ts

core/protect/protect.store.ts
vatsalparikh: Right, makes sense, done, thanks!

core/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
vatsalparikh: Interesting, thanks for pointing this out, done!

core/journey/callbacks/ping-protect-evaluation/ping-protect-evaluation.svelte
vatsalparikh: Interesting, thanks for pointing this out, done!

core/protect/protect.store.ts
SteinGabriel: Awesome, this looks great. Thank you!

cerebrl (approved):

SteinGabriel (approved):
Everything on my end has been addressed and the changes look good. There are only a few files with stale copyright dates, but other than that everything looks great.

Stale copyright dates: 

 - `core/journey/callbacks/ping-protect-initialize/ping-protect-initialize.svelte`
 - `e2e/tests/widget/modal/widget-modal.login.test.js`
 - `packages/login-widget/scripts/copyTypes.mjs`

vatsalparikh:
> Everything on my end has been addressed and the changes look good. There are only a few files with stale copyright dates, but other than that everything looks great.
> 
> Stale copyright dates:
> 
> * `core/journey/callbacks/ping-protect-initialize/ping-protect-initialize.svelte`
> * `e2e/tests/widget/modal/widget-modal.login.test.js`
> * `packages/login-widget/scripts/copyTypes.mjs`

done, thanks!

PR #554, opened 2026-07-20
chore: move-dependencies

ryanbas21:
move deps out of package.json for the widget, allow vite to bundle it all


vatsalparikh (approved):
We were bundling things together in v1.3.0 release https://github.com/ForgeRock/forgerock-web-login-framework/blob/v1.3.0/rollup.config.js.

This PR follows that. And inline dynamic imports need to be removed for code splitting.

All changes in vite.config.ts file make sense.

Approving

